Issue #129, opened 2020-05-19
clonotype_imbalance not found in conda package. 


grst:
- Another thing I encountered was the following, when trying out the full run of the tutorial with the 3k subset of data
```
###################################

>>> ir.pl.clonotype_imbalance(adata, replicate_col='sample', groupby='source', case_label='Tumor', additional_hue='site', plot_type='strip')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'scirpy._plotting' has no attribute 'clonotype_imbalance'
>>> ir.pl.clonotype_imbalance(adata, replicate_col='sample', groupby='source', case_label='Tumor', additional_hue='diagnosis', plot_type='volcano')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'scirpy._plotting' has no attribute 'clonotype_imbalance'

>>> ir.pl.
ir.pl.COLORMAP_EDGES(            ir.pl.clonal_expansion(          ir.pl.embedding(                 ir.pl.spectratype(
ir.pl.alpha_diversity(           ir.pl.clonotype_network(         ir.pl.group_abundance(           ir.pl.vdj_usage(
ir.pl.base                       ir.pl.clonotype_network_igraph(  ir.pl.repertoire_overlap(       


>>> print(ir.__version__)
0.0.0

###################################
```
Looks like the function is not found?
Double checking with conda, the recipe is correctly listed in its version, but is not displayed when printing the info in the console.

grst:
might simply note have been part of the last release. Will make sure it's in the next one.

Issue #128, opened 2020-05-19
Missing tqdm when installing through conda


grst:
I used the conda installation procedure as listed here:

```
###################################

(base) Federicos-MBP:Development fede$ conda create --name playground_scirpy scirpy
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /Users/fede/miniconda3/envs/playground_scirpy

  added / updated specs:
    - scirpy


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    anndata-0.7.1              |           py36_0         155 KB  conda-forge
    blosc-1.18.1               |       h4a8c4bd_0         859 KB  conda-forge
    cython-0.29.17             |   py36h0130604_0         2.1 MB  conda-forge
    glib-2.64.2                |       h577aef8_0         3.2 MB  conda-forge
    h5py-2.10.0                |nompi_py36h106b333_102        1020 KB  conda-forge
    importlib-metadata-1.6.0   |   py36h9f0ad1d_0          42 KB  conda-forge
    importlib_metadata-1.6.0   |                0           3 KB  conda-forge
    kiwisolver-1.2.0           |   py36h863e41a_0          57 KB  conda-forge
    legacy-api-wrap-1.2        |             py_0          18 KB  conda-forge
    libcxx-10.0.0              |       h1af66ff_2         1.0 MB  conda-forge
    llvm-openmp-10.0.0         |       h28b9765_0         268 KB  conda-forge
    llvmlite-0.31.0            |   py36hde82470_1         251 KB  conda-forge
    matplotlib-base-3.2.1      |   py36h83d3ec1_0         7.1 MB  conda-forge
    mock-3.0.5                 |   py36h9f0ad1d_1          44 KB  conda-forge
    nose-1.3.7                 |py36h9f0ad1d_1004         213 KB  conda-forge
    numba-0.48.0               |   py36h4f17bb1_0         3.2 MB  conda-forge
    numexpr-2.7.1              |   py36hcc1bba6_1         128 KB  conda-forge
    openssl-1.1.1g             |       h0b31af3_0         1.9 MB  conda-forge
    packaging-20.1             |             py_0          31 KB  conda-forge
    pandas-1.0.3               |   py36hcc1bba6_1        10.2 MB  conda-forge
    parasail-python-1.2        |   py36hfbae3c0_0         1.9 MB  bioconda
[...]

PR #136, opened 2020-05-22
CI-test the bioconda recipe

grst:
This should prevent issues such as #128, #129. 
Also, this tells us when we have to update dependencies on bioconda.


grst:
Fixes #129. Fixes #128.

Issue #53, opened 2020-04-09
List of improvements for big datasets


grst:
- [ ] speed up group_abundance
- [ ] alignment dist as CLI that reada and writes `anndata` -> to be able to submit this easily to a cluster
- [ ] parallelize chain pairing
- [x] increase chunk size in multiprocessing (less IOwait for pickling... maybe can reduce the amount of data sent to each worker)
- [x] sensible defaults for graph layout. How would force atlase work? 
- [x] increase point size and set edges to False, when plotting many clonotypes
- [x] ~~optimize/parallelize construction and reduction of coord dictionary~~.  (superseeded by #191)
- [x] ~~further optimize matrix housekeeping in `tcr_neighbors`~~ (superseeded by #191)

grst:
Maybe this can be abused to compute Levenshtein distance in linear time: 
https://github.com/wolfgarbe/SymSpell

and/or to prefilter which alignments to compute

grst:
Consider storing clonotype x clonotypes matrices rather than cell x cell matrices. 
This could save a lot of space & computation time
The clonotype network plot would then be updated to show a single dot per 
clonotype (potentially connected into clusters), where the size represents the number of cells.

grst:
I tried to compute edit distance with [symspellpy](https://github.com/mammothb/symspellpy), but is actually 3x slower than the current levenshtein implementation. The reason for this is probably that the Levenshtein package is a highly optimized C library, while symspellpy is a pure python port of the SymSpell library. 
Would need a Python wrapper for C SymSpell for max performance.

grst:
Maybe WFA provides even better alignment performances than parasail? 
https://github.com/smarco/WFA
Need to try it out at some point. In any case, we would need a python wrapper for that library first.

grst:
[mmseqs2](https://github.com/soedinglab/mmseqs2) should do a great job for this. 
Just tried it for some bacterial sequences (many-against-many). It's super fast and I don't see why we couldn't apply it here.

grst:
superseded by #190 (faster chain_pairing) and #230.

PR #142, opened 2020-05-27
Use adata.obsp instead of adata.uns for graphs.

### LINKED ISSUE #133: Use anndata obsp for graphs
The latest version of anndata has improved support for pairwise information (graphs)
and has deprecated storing such information in adata.uns. 

This is possible with AnnData >= 0.7, which we have as a minimal
requirement.

https://github.com/theislab/anndata/commit/be337a8733e8127d96ba4aabac46711fe77e7927

grst:
This is possible with AnnData >= 0.7, which we have as a minimal
requirement.

Closes #133.


grst:
Something here didn't work. 

Moving forward, I instead will not store clonotype connections at a cell-level any more (causes large-memory-inefficient non-sparse blocks when clonotype sizes get large). Instead, we will just store a clonotype x clonotype matrix in `adata.uns`. We already store cell/clonotype membership in `.obs` already anyway. 

The related issue is #53 .

PR #150, opened 2020-06-02
Polish experimental plots

szabogtamas:
- fixed typhos
- removed a heatmap
- added explanations to imbalance plots  
- removed dendrogram
- added gene expression-related examples


grst:
By setting the cell-type in jupyterlab to "raw" and choosing "raw type" = "restructured text" you can use sphinx directives in the notebook, for instance to cite papers (```:cite:`Bibtex-key` ```) or to reference a function (```:func:`scirpy.tl.clonotype_overlap` ```). 

Using ```:func:`~scirpy.tl.clonotype_overlap` ``` will just show the function name, while omitting the `~` will show the entire namespace hierarchy. 

image

docs/tutorials/tutorial_3k_tcr.md
```diff
@@ -507,38 +508,38 @@ ir.pl.spectratype(
...
 
 ```python
+df, dst, lk = ir.tl.repertoire_overlap(adata, "sample", inplace=False)
 df.head()
 ```
 
-```python
-ir.pl.repertoire_overlap(adata, "sample")
-```
+The distance matrix can be shown as a heatmap, while samples are reordered based on the linkage computed previously.
```
grst, line 519: It's not really using the linkage from the cell above (`inplace=False`). How about 
"reordered based on hierarchical clustering"?

docs/tutorials/tutorial_3k_tcr.md
```diff
@@ -507,38 +508,38 @@ ir.pl.spectratype(
...
+A specific pair of samples can be compared on a scatterplot, where dot size corresponds to the number of clonotypes at a given coordinate.
 
 ```python
-ir.pl.repertoire_overlap(adata, "sample", pair_to_plot=["LN2", "LT2"])
+ir.pl.repertoire_overlap(adata, "sample", pair_to_plot=["LN2", "LT2"],fig_kws={"dpi": 120})
 ```
 
+### Discovering clonotypes preferentially occuring in a group
+
+Clonotypes associated with an experimental group (a given cell type, samle or diagnosis) might be important candidates as biomarkers or disease drivers. Scirpy offers `clonotype_imbalance` to rank clonotypes based on Fisher's exact test comparing the fractional presence of a given clonotype in two groups.
```
grst, line 533: This is more for me to understand, but maybe this could be added here as well: 
How would the contingency table look like that you use for fisher's test? 
Is this robust to clonotypes with just a single cell in one sample and no cell in the other?

docs/tutorials/tutorial_3k_tcr.md
```diff
@@ -507,38 +508,38 @@ ir.pl.spectratype(
...
 # Create a column in obs that gives us the site of the tumor
 
 adata.obs["site"] = adata.obs["patient"].str.slice(stop=-1)
 ```
 
 ```python
-# Show clonotypes that are the most imbalance between Tumor and Control samples based on Fischer's test
+# Show clonotypes that are most imbalanced between Tumor and Control samples based on Fisher's test
 
 ir.pl.clonotype_imbalance(
```
grst, line 544: This is probably not for now due to the time constraints... but Is there any way of visualizing this in a more compact way?  Maybe a heatmap? Somehow a lot of space is wasted with a two-axis panel while most of the dots are zero anyway.

grst: Feel free to turn that into an issue for later...

docs/tutorials/tutorial_3k_tcr.md
grst: And can you maybe add a comment what the "normalized abundance" refers to in this case?

docs/tutorials/tutorial_3k_tcr.md
```diff
@@ -560,9 +563,6 @@ ir.pl.clonotype_imbalance(
     case_label="Tumor",
     additional_hue="diagnosis",
     plot_type="volcano",
+    fig_kws={"dpi": 120},
 )
 ```
```
grst, line 568: It would be really nice if you could pick the top "differentially enriched" clonotypes between cell-type annotations and show them next to the cluster labels on the UMAP plot. I think this would help to better understand the concept of differentially enriched. 
Does that make sense? Or am I missing something?

szabogtamas: I've started to do this. Sounds good, but will see after we've tried.

docs/tutorials/tutorial_3k_tcr.md
szabogtamas: Yes, let us leave it as an issue. Heatmap might be a good idea, but might not be necessary before submission

docs/tutorials/tutorial_3k_tcr.md
szabogtamas: Normalized abundance is the number of cells in the simplest case. If cell numbers in samples are very different, then this can be a fraction in a given sample or any other weighted value. But let us say, we stay with the simple case and let it be cell numbers for now and compare tumor samples to a control.
The contingency table would then be: [[nTc, nCc], [sumTc-nTc, sumCc-nCc]], where
- nTc is the number of cells in all Tumor samples that belong to clonotype X
- nCc is the number of cells in all Control samples that belong to clonotype X
- sumTc is the sum of cells in all Tumor samples
- sumCc is the sum of cells in all Control samples  
I would also want to know how robust this approach is. That was one of the reasons I termed this "experimental". The ratios are already very small if we take a look at total cell numbers and cells that belong to a given clonotype.  
But at least theoretically, Fisher's test can handle zero values.

Issue #191, opened 2020-10-02
Refactor CDR3-network construction. 


grst:
Currently, we first build a `sequence` x `sequence` CDR3-similarity network. We then expand this network to a `cell` x `cell` similarity network. 

Using a `cell` x `cell` similarity network is suboptimal. If there are highly expanded clonotypes (i.e. cells with the same CDR3 sequences), this leads to huge blocks in the matrix, that cannot be efficiently stored as a sparse matrix. 

This issue proposes to get rid of the `cell` x `cell` matrix entirely and use the clonotype annotation + CDR3-similarity network directly instead. 

Supersedes #135. (no patching of the edges function will be required any more)

Affected functions: 
 - [ ] Plotting the clonotype network: Instead of blobs of individual cells, we would show connected bubbles. The size of the bubbles corresponds to the clonotype size. Multiple colors can be visualized as a pie chart. 
     * it should be possible to visualize the clonotype number even when not coloring by clonotype
 - [ ] tbc

Additional changes: 
- [ ] Currently, the `same_v_gene` and `same_receptor_type` options only work on a clonotype-definition level. It would be beneficial to offer these already on a network-level, i.e. that nodes with different v-genes are not connected within the network. 
- [ ] will affect how dual_tcr works. Good chance to take care of #164 (make dual_tcr more prominent)

### Notes

 * the trick to make this work is to move the `receptor_arms` and `dual_tcr` arguments to the `define_clonotypes` function, and the clonotype definition *before* the network construction.

josinejansen:
Dear ICBI lab,

I have some questions regarding the Scirpy package, I would really appreciate your help.
For my analyses I first merged TCR and transcriptomics data from two different samples (organ 1 and organ 2), subsequently I then merged these two files.

**1. a) About the clonotype_network: as I understand, each node represents a cell, and edges connect cells belonging to the same clonotype. The function makes visualization of the clonotype-network possible, analogous to the construction of a neighborhood graph from transcriptomics data with the Scanpy package, so that based on the above, it computes a neighborhood graph of CDR3 nucleotide sequences with "scirpy.pp.tcr_neighbors())". My question is: are the clusters (where each node represents a cell) based on identical nucleotide sequences, or is it based on similarity, meaning that one cluster could consists of cells with slightly different nucleotide sequences?**

**b) Do the lines that connect the cells within a cluster have a meaning?**

[...]

PR #216, opened 2020-11-30
Immunologist-approved BCR similarity measure implementation attempt

ktpolanski:
Figured that rather than make an issue asking for a thing I'd tried to implement the thing myself, to mixed results.

I have consulted with an immunologist with regards to BCR clonotype calling, and apparently somatic hypermutation only leads to substitution events. As such, an appropriate way to identify neighbours would be to require sequence length compatibility, and then use Hamming distance to identify substitutions. Going into your source code, the logic for the Levenshtein distance was clear:

        result = []
        for (row, s1), (col, s2) in coord_iterator:
            d = levenshtein_dist(s1, s2)
            if d <= self.cutoff:
                result.append((d + 1, origin_row + row, origin_col + col))

As such, I replicated it for the Hamming distance plus sequence length, and added references to it wherever Levenshtein cropped up:

        result = []
        for (row, s1), (col, s2) in coord_iterator:

            # require identical length of sequences
            if len(s1) != len(s2):
                continue
            d = hamming_dist(s1, s2)
            if d <= self.cutoff:
                result.append((d + 1, origin_row + row, origin_col + col))

Good news is this seems to work, kind of. It produces a graph and everything, and the graph balls seem sensible, but they also appear to have no edges.

image

Also, for whatever reason, it takes longer to run than the Levenshtein, with higher process count numbers during the progress bars. This seems weird, as if anything you'd imagine the metric would be more discerning?

```
100%|██████████| 561/561 [00:01<00:00, 313.37it/s]
100%|██████████| 990/990 [00:03<00:00, 320.64it/s]
100%|██████████| 14098/14098 [00:00<00:00, 55333.33it/s]
100%|██████████| 22884/22884 [00:00<00:00, 47773.14it/s]
100%|██████████| 122761/122761 [00:01<00:00, 105207.93it/s]
... storing 'ct_cluster_nt_levenshtein' as categorical
```

```
100%|██████████| 561/561 [00:02<00:00, 196.40it/s]
100%|██████████| 990/990 [00:04<00:00, 206.99it/s]
100%|██████████| 218911/218911 [00:04<00:00, 52849.82it/s]
100%|██████████| 406018/406018 [00:08<00:00, 47138.79it/s]
100%|██████████| 1903399/1903399 [00:17<00:00, 106348.03it/s]
... storing 'ct_cluster_nt_hamming-substitution' as categorical
```

Maybe my attempt at an implementation will be of use, maybe it won't. The general idea still stands - a neighbour graph that requires sequence length identity and then computes Hamming is immunologist-blessed for BCR processing, and would be nice as an option.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -21,21 +21,21 @@ jobs:
           fetch-depth: 0 # required for setuptools-scm
 
       - name: Setup Minoconda
-        uses: goanpeca/setup-miniconda@v1
+        uses: goanpeca/setup-miniconda@v2
         with:
           auto-update-conda: true
+          mamba-version: "*"
+          channels: conda-forge,bioconda,defaults
+          channel-priority: true
           python-version: ${{ matrix.python-version }}
 
-      - name: Set-up channels and install conda build
+      - name: install conda build
         run: |
-          conda config --add channels defaults
-          conda config --add channels bioconda
-          conda config --add channels conda-forge
-          conda install -y conda-build conda-verify
+          mamba install -y conda-build conda-verify
         shell: bash
 
       - name: build and test package
         run: |
           cd .conda
-          conda build --no-anaconda-upload .
+          mamba build --no-anaconda-upload .
         shell: bash
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -20,8 +20,8 @@ jobs:
         with:
           fetch-depth: 0 # required for setuptools-scm
 
-      - name: Setup Minoconda
-        uses: goanpeca/setup-miniconda@v1
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
           mamba-version: "*"
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -21,7 +21,7 @@ jobs:
           fetch-depth: 0 # required for setuptools-scm
 
       - name: Setup Minoconda
-        uses: goanpeca/setup-miniconda@v2
+        uses: goanpeca/setup-miniconda@v1
         with:
           auto-update-conda: true
           mamba-version: "*"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -82,7 +82,7 @@ jobs:
         if: ( matrix.os == 'ubuntu-latest' ) &&  ( matrix.python-version == '3.7' )
         uses: JamesIves/github-pages-deploy-action@releases/v3
         with:
-          ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
           BRANCH: gh-pages # The branch the action should deploy to.
           FOLDER: docs/_build/html # The folder the action should deploy.
           TARGET_FOLDER: ${{ env.target_dir }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -74,7 +74,7 @@ jobs:
           # when on master, we want to deploy to the root dir (-> ".")
           target_dir=$(echo $tmp_target_dir | sed "s/heads\/master/./")
           echo $target_dir
-          echo "::set-env name=target_dir::$target_dir"
+          echo "target_dir=$target_dir" >> $GITHUB_ENV
         env:
           github_ref: ${{ github.ref }}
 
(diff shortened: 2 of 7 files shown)


grst:
Hi @ktpolanski, 

first, thanks a lot for this PR. This is the first public code contribution to scirpy, and I really appreciate it :tada: 

---

> As such, an appropriate way to identify neighbours would be to require sequence length compatibility, and then use Hamming distance to identify substitutions. 

Just to make sure I understood it correctly: you define the `hamming-substitution` metric as 

 * `0` (=`inf`) -> no edge, iff the CDR3 sequence is of different length, and
 * number of substitutions, iff the CDR3 sequences are of the same length? 

I would also be curious to know what @szabogtamas thinks of this metric :) 

---

>  It produces a graph and everything, and the graph balls seem sensible, but they also appear to have no edges.

Edge plotting is disabled for > 1000 nodes for performance reasons. You can force edges using the `edges=True` argument. 

---

> Also, for whatever reason, it takes longer to run than the Levenshtein, with higher process count numbers during the progress bars. This seems weird, as if anything you'd imagine the metric would be more discerning?

 * The first two progress bars are limited by the number of unique CDR3 sequences. Even here the hamming distance seems slightly slower, but this may very well be the case as the `Levenshtein` package uses a highly optimized C implementation. It also has a `hamming` function (see [here](https://rawgit.com/ztane/python-Levenshtein/master/docs/Levenshtein.html)), maybe it is slightly faster.
* The last three progress bars are limited by the number of edges in the graph. Especially large groups of cells fully connected by edges can slow this down. I need to refactor this part at some point to make it more efficient (https://github.com/icbi-lab/scirpy/issues/191), but it's nothing to worry about in this PR. 


Best, 
Gregor

ktpolanski:
Thanks for the feedback. Yep, that's how this works. The biological motivation is that as per @zktuong, somatic hypermutation only leads to substitutions.

I switched the distance measure to the Levenshtein module Hamming, and suddenly everything runs instantly, has smaller process numbers, and the graphs look different/nicer. Huh, weird. I have no idea why this would be the case, but that's what happened.

When replicating your Levenshtein functionality, I copied over the `d+1` aspect of it. Is there any particular reason why you do this by the way?

Also, you'd think that running this with `cutoff=0` would produce the identity graph in a roundabout way, but instead it just fails to create a graph instead. This is also the case for Levenshtein. Huh.

Also also, I promise I just changed the Hamming-related stuff (as you can see in the commits anyway), I have no idea why the tests are crashing out, sorry.

grst:
> I switched the distance measure to the Levenshtein module Hamming, and suddenly everything runs instantly, has smaller process numbers, and the graphs look different/nicer. Huh, weird. I have no idea why this would be the case, but that's what happened.

I just had a look, apparently `scipy.spatial.distance.hamming` returns a normalized value between 0 and 1 (not what I expected, either!) 

> When replicating your Levenshtein functionality, I copied over the d+1 aspect of it. Is there any particular reason why you do this by the way?

This is required to efficiently represent the graph as a sparse matrix. All edges with a distance `> cutoff` need to be 0, therefore, the distance needs to be offset by 1. 

> Also also, I promise I just changed the Hamming-related stuff (as you can see in the commits anyway), I have no idea why the tests are crashing out, sorry.

The first problem I see is that you need to format your code using [black](https://github.com/psf/black). 
There seems to be another problem with setting up the environment with some Python versions. That's not your fault and I'll look into it.

ktpolanski:
I put `HammingSubstitutionDistanceCalculator()` into the [black converter](https://black.now.sh/) and it didn't change anything, so that part seems to be okay.

grst:
image

ktpolanski:
Thanks for the pro tip, manually  ran `black` on the two files in question, and have now set up the `pre-commit` thing.

grst:
Great!
Would you mind adding a small test case here, similar to the levenshtein one? 
https://github.com/icbi-lab/scirpy/blob/6e975f197131024737d798f715a363112c06e053/scirpy/tests/test_tcr_dist.py#L162-L172

grst:
How about calling the distance just `hamming`, for brevity? 
Other than that this looks good now, and I'll fix the remaining tests.

grst:
oh boy, why is CI always so painful :roll_eyes: 

I'll merge this asap.

ktpolanski:
Fingers crossed, seems like this did the trick as it's yet to fail!

Thanks for the assistance/guidance.

grst:
ok, finally merged! 

I'll fix a few other small things and then make a release, probably at some point next week. 

Thanks again for your contribution!

ktpolanski:
Cool, thanks a lot! :)

grst:
now available as `v0.6.0`.

Issue #132, opened 2020-05-19
Support for BCR


grst:
# BCR support meta issue. 

## Initial PR #183 addresses: 
- [x] change datastructure

 * Instead of `TRA_1`, `TRA_2`, `TRB_1` and `TRB_2` have `arm1_primary`, `arm1_secondary`, `arm2_primary` and `arm2_secondary`
 * have additional 4 columns: `arm1_primary_type`, ...; These can accept values such as `TRA`, `TRB`, `TRG`, `TRD`, `IGH`, `IGL` (see [AIRR locus names](https://docs.airr-community.org/en/latest/datarep/rearrangements.html#locus-names))

 - [x] adapt chain_categories to identify *bona fide* vs other pairings (e.g. flag a cell with TRA + IGH). 
 - [x] clonotype network: separate by receptor_type (per default, no connections between BCR and TCR)
 - [x] update glossary and documentation. Make clear that now there are `VJ` and `VDJ` chains. (mostly done)
 - [x] rename `tcr_` to `cdr3_` or `receptor_` or `vdj_` or `ir_`

## To be resolved before next release (v0.5 "adding experimental BCR support")
 - [x] #194 vdj_usage is broken with new data structure
 - [x] #195 improve BCR-related documentation
 - [x] `read_tracer()` with gamma/delta
 - [x] `read_bracer()` 
 - [x] #198 Add BCR example dataset

## BCR-related issues that can be resolved at a later point
 - [ ] #197 function to infer antibody class
 - [ ] #196 function to infer somatic hypermutation status
 - [ ] #199 BCR-tutorial
 - [ ] add support for CDR1 and 2 (#185)

ktpolanski:
Leaving a line here in case public support can help prioritise development. VDJ analysis is not quite as standardised, it would be great to have a single well-documented package to handle both TCRs and BCRs.

grst:
Hi @ktpolanski,

thanks for your comment! BCR support is definitely THE next thing on the list, atm I'm just a bit overloaded with other tasks. 

Can you, by any chance, recommend a good (publicly available) example dataset with single-cell BCR data that we could use to build the tutorial? 

Cheers, 
Gregor

grst:
@szabogtamas, @FFinotello, @ktpolanski 

Do you guys have any suggestions on how to name the columns for BCR support? 

At the moment we have: 
 * TRA_1_<attr>/TRA_2_<attr>: columns related to the primary and secondary TCR-alpha chain
[...]

PR #241, opened 2021-02-17
Improve interoperability by fully supporting the AIRR standard

### LINKED ISSUE #215: Repertoire overlap seemingly invariant to clonotype network changes
Hello,

I just tried analysing some VDJ data with the package, and encountered some unusual behaviour. I mimicked the tutorial's syntax for calling clonotypes via nucleotide identity and amino acid alignment, and tried to use the results to assess repertoire overlap.

```
ir.pp.ir_neighbors(adata)
ir.tl.define_clonotypes(adata)
ir.tl.clonotype_network(adata, min_size=2)
ir.pl.clonotype_network(adata, color="Donor")
ir.tl.repertoire_overlap(adata, "ind_loc")
ir.pl.repertoire_overlap(adata, "ind_loc", heatmap_cats=["Donor", "Spatial_code"])

ir.pp.ir_neighbors(
    adata,
    metric="alignment",
    sequence="aa",
    cutoff=15,
    receptor_arms="all",
    dual_ir="all",
)
ir.tl.define_clonotype_clusters(
    adata, partitions="connected", sequence="aa", metric="alignment", within_group=None
)
ir.tl.clonotype_network(adata, min_size=4, sequence="aa", metric="alignment")
ir.pl.clonotype_network(adata, color="Donor")
ir.tl.repertoire_overlap(adata, "ind_loc")
ir.pl.repertoire_overlap(adata, "ind_loc", heatmap_cats=["Donor", "Spatial_code"])
```

The two networks come out looking quite different:

image
image

However, in both cases, I get an identical repertoire overlap plot, seemingly derived from the first one?

image

I can't seem to find any parameters controlling `repertoire_overlap()` in a manner analogous to `clonotype_network()`. According to the documentation entry for `define_clonotypes()` and `define_clonotype_clusters()`, they both save their results to  `.obs['clonotype']`, which is `repertoire_overlap()`'s default input. As such, I'm stumped. Some assistance please?

### LINKED ISSUE #242: Label filtered contig annotations?
When calling `scirpy.io.read_10x_vdj()` with `filtered=False` it might be useful to add a categorical attribute like `adata.obs["has_ir_filtered"]` to distinguish those cells (or perhaps better use the original `is_cell` and `high_confidence` labels)

### LINKED ISSUE #253: IrChain instantiation fails when no `expr` parameter is specified
Thanks for this amazing toolkit!
I am trying to import my data in "other format" in a similar fashion as in the [importing tutorial](https://icbi-lab.github.io/scirpy/tutorials/tutorial_io.html?highlight=adaptive%20immune%20receptor#Creating-AnnData-objects-from-other-formats). However, my data does not have a TCR expression field, so when I try to omit it, it takes the default value (`expr=None`) and this crashes an internal casting operation. In this toy example you can see the error:
```
import numpy as np
import pandas as pd
import scanpy as sc
import scirpy as ir
from matplotlib import pyplot as plt

ir.io.IrChain(locus='TRA',cdr3_nt='CCACACTTTGTCA')
```
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-c503e025a0de> in <module>
----> 1 ir.io.IrChain(locus='TRA',cdr3_nt='CCACACTTTGTCA')

~/anaconda3/lib/python3.7/site-packages/scirpy/io/_datastructures.py in __init__(self, locus, cdr3, cdr3_nt, expr, expr_raw, is_productive, v_gene, d_gene, j_gene, c_gene, junction_ins)
     95         self.cdr3 = cdr3.upper() if not _is_na(cdr3) else None
     96         self.cdr3_nt = cdr3_nt.upper() if not _is_na(cdr3_nt) else None
---> 97         self.expr = float(expr)
     98         self.expr_raw = float(expr_raw) if not _is_na(expr_raw) else None
     99         self.is_productive = _is_true(is_productive)

TypeError: float() argument must be a string or a number, not 'NoneType'
```
As you can see I am using anaconda3, python 3.7 in a jupyter notebook with Ubuntu 20.04. The versions of the modules are the following:
```
sc.logging.print_header()
scanpy==1.7.1 anndata==0.7.5 umap==0.4.6 numpy==1.17.1 scipy==1.5.2 pandas==1.2.3 scikit-learn==0.23.2 statsmodels==0.11.1 python-igraph==0.9.0
```
```
ir.__version__
'0.6.1'
```

### LINKED ISSUE #255: Replace `adata.concatenate` with `anndata.concat` in the tutorial
`concatenate` breaks when no transcriptomics data is available, see https://github.com/theislab/anndata/issues/526

### LINKED ISSUE #257: Can't save anndata after running `define_clonotypes`
reported by @chris-rands in #243: 

> * I can't save the object to .h5ad due to `TypeError: Can't implicitly convert non-string objects to strings` with `Above error raised while writing key 'uns/clonotype/cell_indices' of <class 'h5py._hl.files.File'> from /.` I encountered this with parts of scanpy also, e.g. [theislab/scanpy#1651 (comment)](https://github.com/theislab/scanpy/issues/1651#issuecomment-779382361)

### LINKED ISSUE #258: how to populate IR_*_clone_id columns?
Hi @grst, I can't work out when in the analysis workflow is the `IR_*_clone_id` columns populated. Toggling `dual_ir` option doesn't seem to change that behaviour. Does it actually get populated from preprocessing/analysis performed within scirpy?

grst:
See also https://github.com/icbi-lab/scirpy/issues/132#issuecomment-776513292 and https://github.com/icbi-lab/scirpy/issues/132#issuecomment-775861882

- [x]  Rename IrCell and IrChain to AirrCell and AirrChain to reflect that they are fully AIRR compliant
- [x]  Change them to support all columns defined in the rearrangement schema. Except the fields required by the AIRR standard  all fields are optional.
- [x] I found a way ~~(although a hacky one: pickling and base64 encoding)~~ (json serialization) of storing excessive chains in an extra chain. They would not be used by scirpy, but it would make reading AIRR files non-destructive.
- [x] Implement a write_airr function in addition to the {to,from}_dandelion functions you are working on already.
- [x] Integrate changes from #240 
- [x] Consider renaming `IR_xxx` columns to match the AIRR schema
- [x] Include additional 10x columns (#242 )
- [x] https://github.com/icbi-lab/scirpy/issues/253: Tolerate missing columns
- [x] Rename `clonotype` column to `clone_id` (#258)
- [x] 3k tutorial 
- [ ] Update data loading tutorial (#255)
- [ ] Go through entire documentation (API and pages), in particular page about data structure. 
- [x] Print out where clonotype information is stored (https://github.com/icbi-lab/scirpy/issues/215#issuecomment-722984151)
- [x] Add test-case for #254 with new merge functions
- [x] #257 (save anndata after define_clonotypes) 

Closes #253
Closes #258
Closes #255 
Closes #242
Closes #215
Closes #257


scirpy/io/_convert_anndata.py
```diff
@@ -174,28 +133,37 @@ def to_ir_objs(adata: AnnData) -> List[IrCell]:
...
+            cell_id = ix.cell_id
+            sequence_id = cell_id + "_contig_" + str(counter)
+            contig_dict.update(
+                {
+                    "cell_id": cell_id,
+                    "sequence_id": sequence_id,
+                    "locus": c.locus,
+                    "junction_aa": c.cdr3,
+                    "junction": c.cdr3_nt,
+                    "umi_count": c.expr,
```
grst, line 158: @zktuong, should `umi_count` be `duplicate_count`?

zktuong: To be 100% compliant with AIRR, yes. Dandelion allows both and will check for presence of `umi_count` first before checking for `duplicate_count` but i can/should update this to just stick to duplicate_count.

grst: Ok, so it doesn't hurt to use `duplicate_count` here in any case. 

How common is the `umi_count` column anyway? I was considering not to support it as it makes stuff more complicated and people could always manually rename it to `duplicate_count` if they have a non-compliant file.

zktuong: I think that `umi_count` is more intuitive than `duplicate_count` but that personal preference. I think the immcantation tools parse it as `umi_count` whereas the 10x AIRR table (and rightly so) is provided as `duplicate_count`. Most people would still be using immcantation-processed files so there would certainly be interest in easy porting between formats but like you said, they can just change their column name first before trying and this will free up time and effort on your end.

scirpy/io/_io.py
```diff
@@ -392,54 +434,135 @@ def read_bracer(path: Union[str, Path]) -> AnnData:
...
+    # replace "unassigned" with None
+    for col in dandelion_df.columns:
+        dandelion_df.loc[dandelion_df[col] == "unassigned", col] = None
+
+    adata = read_airr(dandelion_df)
+    if transfer:
+        ddl.tl.transfer(
+            adata, dandelion
+        )  # need to make a version that is not so verbose?
+    return adata
```
grst, line 568: @zktuong, most of this PR is still work-in-progress, but the round trip-conversion of a 10x dataset from scirpy to dandelion and back seems to work now. Could you please have another look at these functions if everything makes sense? 

Unlike the previous `IrCell/IrChain` implementation, `AirrCell` now simply stores chains as d AIRR rearrangement-compliant dictionaries.

zktuong: A couple of hiccups. I've attached the notebook as a pdf.

[testing_scirpy_dandelion_conversion.pdf](https://github.com/icbi-lab/scirpy/files/6228705/testing_scirpy_dandelion_conversion.pdf)

One of it is on my end -> I need to switch the requirement for umi_count to allow duplicate_count as mentioned. Seems like my init method is not catching that properly.

zktuong: also it's overly verbose i think

zktuong: ah actually the issue is with 'locus'. it's doesn't come back as a column with `read_airr` (or `airr.load_rearrangement`), although it's a standard term (but optional according to airr). It's a hard requirement for dandelion.

grst: Thanks for having a look! In your PR you wrote a function to infer the locus name from the genes (see https://github.com/icbi-lab/scirpy/pull/241/files#diff-5e72138c27e2f83f0219586fad1d67885a1b16ceaf3e327447bba0f5ab1f85d9R379). That's certainly a useful feature and I'm adding it to `read_airr` now.

zktuong: Thanks @grst. That fixed it.

I encountered a weird bug that i will try to fix on dandelion's side first. This has got to do with empty `clone_id` columns in the airr table. scirpy passes that empty entry into `AirrCell` fine but then this messes with dandelion's initialization procedure but a simple fix will correct that. 

It did also revealed that if i were to call clones from scirpy and try to convert to dandelion, this information is not passed on and meets the same issue above. Is it because in scirpy `clone_id` is called `clonotype`?

grst: Currently only the chain-specific attributes get exported (i.e. all scirpy columns that start with `IR_`). 
In principle, it probably makes sense to write out the `clonotype` column to `clone_id`. But then again, scirpy allows 
different versions of clonal assignments...

Maybe something for the next PR :shrug:

zktuong: OK. sounds good. I will just add a small conversion on dandelion's io from scirpy to parse the `clonotype` column to `clone_id` as a default, unless otherwise specified which the user can adjust.

grst: See #258

grst:
@zktuong, pretty much there now, if you want to give it another try. 
I also decided to default to `clone_id` in `define_clonotypes`. 

The remaining changes should be mostly related to updating the documentation.

zktuong:
Yes the round trip back and forth works cleanly regardless of how the airr table is read and processed in dandelion or scirpy! Will update my [notebook](https://sc-dandelion.readthedocs.io/en/latest/notebooks/1c_dandelion_scirpy.html) once it passes the tests.

grst:
documentation update in follow-up PR

PR #283, opened 2021-07-15
adjusting add_serialized_chains behaviour

zktuong:
Hi @grst,

I'm adding some updates to dandelion's parser and came across a small issue and wondering if you could give me a hand.
I've made a small snippet of the all_contig_annotations.json file from 10x's resource to use as a fixture (attached as 'data.txt' here because github doesn't allow me to attach with .json extension):
[data.txt](https://github.com/icbi-lab/scirpy/files/6823960/data.txt)

The following round tripping works:
```python
import dandelion as ddl
import scirpy as ir

adata = ir.io.read_10x_vdj('data.json')
vdj = ddl.read_10x_vdj('data.json') # I've wrote a new parser for dandelion

vdj_1 = ir.io.to_dandelion(adata)
adata_1 = ddl.to_scirpy(vdj) # this basically calls ir.io.from_dandelion

adata_2 = ir.io.from_dandelion(vdj)
vdj_2 = ddl.from_scirpy(adata) # this basically calls ir.io.to_dandelion
```

And while this works:
```python
adata_3 = ddl.to_scirpy(vdj_2)
```

the following doesn't:
```python
vdj_3 = ddl.from_scirpy(adata_3)
adata_4 = ir.io.to_dandelion(adata_2)
```

Both come up with the same issue:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-15-00d75a1a5f48> in <module>
     26     try:
     27         print(row['extra_chains'])
---> 28         tmp_ir_cell.add_serialized_chains(row["extra_chains"])
     29     except KeyError:
     30         print(row['extra_chains'])

~/Documents/Github/scirpy/scirpy/io/_datastructures.py in add_serialized_chains(self, serialized_chains)
    164             tmp_chains = json.loads(serialized_chains)
    165             for chain in tmp_chains:
--> 166                 self.add_chain(chain)
    167 
    168     def _split_chains(self) -> Tuple[bool, dict]:

~/Documents/Github/scirpy/scirpy/io/_datastructures.py in add_chain(self, chain)
    144             self._chain_fields = list(chain.keys())
    145         elif self._chain_fields != list(chain.keys()):
--> 146             raise ValueError("All chains must have the same fields!")
    147 
    148         if "locus" not in chain:

ValueError: All chains must have the same fields!
```

If I add `print(tmp_chain)` in the loop before `self.add_serialized_chains`, the output is this:
```python
{'sequence_id': None, 'sequence': None, 'rev_comp': None, 'productive': 'True', 'v_call': 'IGLV3-1', 'd_call': nan, 'j_call': 'IGLJ2', 'sequence_alignment': None, 'germline_alignment': None, 'junction': 'TGTCAGGCGTGGGACAGCAGCACTGTGGTATTC', 'junction_aa': 'CQAWDSSTVVF', 'v_cigar': None, 'd_cigar': None, 'j_cigar': None, 'c_call': 'IGLC2', 'consensus_count': 464.0, 'duplicate_count': 16.0, 'locus': 'IGL'}
{'sequence_id': None, 'sequence': None, 'rev_comp': None, 'productive': 'None', 'v_call': nan, 'd_call': nan, 'j_call': nan, 'sequence_alignment': None, 'germline_alignment': None, 'junction': nan, 'junction_aa': nan, 'v_cigar': None, 'd_cigar': None, 'j_cigar': None, 'c_call': nan, 'consensus_count': nan, 'duplicate_count': nan, 'locus': nan}
{'sequence_id': None, 'sequence': None, 'rev_comp': None, 'productive': 'True', 'v_call': 'IGHV4-39', 'd_call': 'IGHD3-22', 'j_call': 'IGHJ4', 'sequence_alignment': None, 'germline_alignment': None, 'junction': 'TGTGCGGTCCAGTATTACTATGATAGTAGTGGTTACCGGAGAACTGAGGTGCCCTTTGACTACTGG', 'junction_aa': 'CAVQYYYDSSGYRRTEVPFDYW', 'v_cigar': None, 'd_cigar': None, 'j_cigar': None, 'c_call': 'IGHM', 'consensus_count': 488.0, 'duplicate_count': 30.0, 'locus': 'IGH'}
{'sequence_id': None, 'sequence': None, 'rev_comp': None, 'productive': 'None', 'v_call': nan, 'd_call': nan, 'j_call': nan, 'sequence_alignment': None, 'germline_alignment': None, 'junction': nan, 'junction_aa': nan, 'v_cigar': None, 'd_cigar': None, 'j_cigar': None, 'c_call': nan, 'consensus_count': nan, 'duplicate_count': nan, 'locus': nan}
```

Looking at the json serialised entries in `extra_chains`:
```python
import json
json.loads(row["extra_chains"])
[{'c_call': 'IGKC',
  'consensus_count': 120.0,
  'd_call': None,
  'd_cigar': None,
  'duplicate_count': 3.0,
  'germline_alignment': None,
  'j_call': 'IGKJ1',
  'j_cigar': None,
  'junction': None,
  'junction_aa': None,
  'locus': 'IGK',
  'np1_length': None,
  'np2_length': None,
  'productive': False,
  'rev_comp': None,
  'sequence': None,
  'sequence_alignment': None,
  'sequence_id': 'AAACCTGAGACTGTAA-1_contig_3',
  'v_call': None,
  'v_cigar': None}]
```

It looks like there a few additional keys (`np1_length` and `np2_length`). 

My proposed adjustments solves the round tripping issue but I wasn't sure if it was fitting with how the original class was supposed to function in situations like this were there's a mismatch in the type of fields present in the main chains vs extra chains.

### DIFF
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -163,8 +163,6 @@ class AirrCell(MutableMapping):
         if not _is_na2(serialized_chains):
             tmp_chains = json.loads(serialized_chains)
             for chain in tmp_chains:
-                # keep only keys present in _chain_fields
-                # chain = {k: v for k, v in chain.items() if k in self._chain_fields}
                 self.add_chain(chain)
 
     def _split_chains(self) -> Tuple[bool, dict]:
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -236,17 +236,26 @@ class AirrCell(MutableMapping):
         return tuple(-1 if x is None else x for x in sort_tuple)
 
     @staticmethod
-    def _serialize_chains(chains: List[MutableMapping]) -> str:
+    def _serialize_chains(chains: List[MutableMapping], include_fields: Optional[Collection[str]] = None) -> str:
         """Serialize chains into a JSON object. This is useful for storing
         an arbitrary number of extra chains in a single column of a dataframe."""
         # convert numpy dtypes to python types
         # https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types
         for chain in chains:
             for k, v in chain.items():
-                try:
-                    chain[k] = chain[k].item()
-                except AttributeError:
-                    pass
+                if include_fields is not None:
+                    if k in include_fields:
+                        try:
+                            chain[k] = chain[k].item()
+                        except AttributeError:
+                            pass
+                    else:
+                        pass
+                else:
+                    try:
+                        chain[k] = chain[k].item()
+                    except AttributeError:
+                        pass
         return json.dumps(chains)
 
     def to_airr_records(self) -> Iterable[dict]:
@@ -294,7 +303,7 @@ class AirrCell(MutableMapping):
         include_fields.add("cell_id")
 
         res_dict["multi_chain"], chain_dict = self._split_chains()
-        res_dict["extra_chains"] = self._serialize_chains(chain_dict.pop("extra"))
+        res_dict["extra_chains"] = self._serialize_chains(chain_dict.pop("extra"), include_fields = include_fields)
 
         # add cell-level attributes
         for key in self:
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -163,6 +163,8 @@ class AirrCell(MutableMapping):
         if not _is_na2(serialized_chains):
             tmp_chains = json.loads(serialized_chains)
             for chain in tmp_chains:
+                # keep only keys present in _chain_fields
+                # chain = {k: v for k, v in chain.items() if k in self._chain_fields}
                 self.add_chain(chain)
 
     def _split_chains(self) -> Tuple[bool, dict]:
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -163,7 +163,9 @@ class AirrCell(MutableMapping):
         if not _is_na2(serialized_chains):
             tmp_chains = json.loads(serialized_chains)
             for chain in tmp_chains:
-                self.add_chain(chain)
+                tmp_chain = AirrCell.empty_chain_dict()
+                tmp_chain.update(chain)
+                self.add_chain(tmp_chain)
 
     def _split_chains(self) -> Tuple[bool, dict]:
         """
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -236,7 +236,9 @@ class AirrCell(MutableMapping):
         return tuple(-1 if x is None else x for x in sort_tuple)
 
     @staticmethod
-    def _serialize_chains(chains: List[MutableMapping], include_fields: Optional[Collection[str]] = None) -> str:
+    def _serialize_chains(
+        chains: List[MutableMapping], include_fields: Optional[Collection[str]] = None
+    ) -> str:
         """Serialize chains into a JSON object. This is useful for storing
         an arbitrary number of extra chains in a single column of a dataframe."""
         # convert numpy dtypes to python types
@@ -303,7 +305,9 @@ class AirrCell(MutableMapping):
         include_fields.add("cell_id")
 
         res_dict["multi_chain"], chain_dict = self._split_chains()
-        res_dict["extra_chains"] = self._serialize_chains(chain_dict.pop("extra"), include_fields = include_fields)
+        res_dict["extra_chains"] = self._serialize_chains(
+            chain_dict.pop("extra"), include_fields=include_fields
+        )
 
         # add cell-level attributes
         for key in self:
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -251,8 +251,6 @@ class AirrCell(MutableMapping):
                             chain[k] = chain[k].item()
                         except AttributeError:
                             pass
-                    else:
-                        pass
                 else:
                     try:
                         chain[k] = chain[k].item()
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -163,6 +163,8 @@ class AirrCell(MutableMapping):
         if not _is_na2(serialized_chains):
             tmp_chains = json.loads(serialized_chains)
             for chain in tmp_chains:
+                # keep only keys present in _chain_fields
+                chain = {k: v for k, v in chain.items() if k in self._chain_fields}
                 self.add_chain(chain)
 
     def _split_chains(self) -> Tuple[bool, dict]:
--- a/scirpy/io/_datastructures.py
+++ b/scirpy/io/_datastructures.py
@@ -243,18 +243,17 @@ class AirrCell(MutableMapping):
         # https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types
         for chain in chains:
             for k, v in chain.items():
-                if include_fields is not None:
-                    if k in include_fields:
-                        try:
-                            chain[k] = chain[k].item()
-                        except AttributeError:
-                            pass
-                else:
-                    try:
-                        chain[k] = chain[k].item()
-                    except AttributeError:
-                        pass
-        return json.dumps(chains)
+                try:
+                    chain[k] = chain[k].item()
+                except AttributeError:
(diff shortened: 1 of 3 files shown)


grst:
Hi Kelvin, 

thanks for that detailed report. 

>  how the original class was supposed to function in situations like this were there's a mismatch in the type of fields present in the main chains vs extra chains.

The class expects no differences in the types of fields to be present. Not because this would be strictly required, but rather because it's usually a sign that something went wrong at some point. 

Do you have any idea where these additional `np1_length` fields come from all of a sudden? 
(Or rather: why do they get removed in the 'normal' chains?)

zktuong:
> Do you have any idea where these additional `np1_length` fields come from all of a sudden?
> (Or rather: why do they get removed in the 'normal' chains?)

So I think the issue lies in:
```
adata_3 = ddl.to_scirpy(vdj_2) # or adata_3 = ir.io.from_dandelion(vdj_2) 
```

Which does this: 

```python
def from_dandelion(dandelion, transfer=False) -> AnnData:
   # skipping
    adata = read_airr(dandelion_df)
   # skipping
```

I suspect the issue might be with the `include_fields` option that trims the transferred items to whatever's in `DEFAULT_AIRR_FIELDS`, which doesn't include the `np1_length` etc. keys. 

```python
def read_airr(
    path: Union[str, Sequence[str], Path, Sequence[Path]],
    use_umi_count_col: Union[bool, Literal["auto"]] = "auto",
    infer_locus: bool = True,
    cell_attributes: Collection[str] = DEFAULT_AIRR_CELL_ATTRIBUTES,
    include_fields: Optional[Collection[str]] = DEFAULT_AIRR_FIELDS,
) -> AnnData:
# rest of code
    return from_airr_cells(airr_cells.values(), include_fields=include_fields)
```

Will test a bit and report back.

grst:
That's probably it! 

However, shouldn't the extra chains get converted back to "normal chains" during `to_dandelion`? And then, during `from_dandelion` they should be treated and trimmed like any other chain. 

I'm missing something :thinking:

zktuong:
Yes i think that solved it.

> However, shouldn't the extra chains get converted back to "normal chains" during `to_dandelion`? And then, during `from_dandelion` they should be treated and trimmed like any other chain.

Yea i thought so too. What it looks like is the extra chains and normal chains were extracted differently during the ` `from_airr_cells` bit.

grst:
Ok, I looked into it using your example data. The `np1_length` should never end up in `extra_chains` when using `from_dandelion` (which calls `read_airr` internally). 

It turns out I forgot to filter `extra_chains` for `include_fields` entirely. I think the solution is to pass `include_fields` to the `AirrCell._serialize_chains` function: 
https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L295

Do you want to give it a go?

zktuong:
No dice. I got:
```python
_serialize_chains() got an unexpected keyword argument 'include_fields'
```

grst:
Ah, I meant implementing support for `include_fields` in `_serialize_chains` -- should be trivial. Should have been more specific.

zktuong:
Ah ok sure.

I'm trying to work this out, unfortunately with much difficulty.

The `AirrCell.add_chain` is called by `_serialize_chains` and this works cleanly when dealing with extra or normal chains.
`from_airr_cells`'s `.to_scirpy_record()` works perfectly too. Only thing is it calls this in the first step:
https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_convert_anndata.py#L76-L78

I wonder if this is where it needs to be tweaked?
https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L303-L314

grst:
`_to_scirpy_records` first decides which chains will be extra and which will be "normal": 

https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L294

Then the 'extra' chains get serialized and added: 
https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L295

So ultimately, `_serialize_chains` would need to be modified to accept an `include_fields` parameters:
https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L236-L248

The section you mentioned above already handles `include_fields` correctly for 'normal' chains. However, by then the `extra_chains` already got added including all fields.

zktuong (commented):
something like this?

scirpy/io/_datastructures.py
```diff
@@ -234,17 +236,28 @@ def _key_sort_chains(chain) -> Tuple:
...
-                except AttributeError:
-                    pass
+                if include_fields is not None:
+                    if k in include_fields:
+                        try:
+                            chain[k] = chain[k].item()
+                        except AttributeError:
+                            pass
+                    else:
+                        pass
```
grst, line 255:
```suggestion
```

scirpy/io/_io.py
```diff
@@ -663,7 +663,9 @@ def to_dandelion(adata: AnnData):
 
 
 @_doc_params(doc_working_model=doc_working_model)
-def from_dandelion(dandelion, transfer=False) -> AnnData:
+def from_dandelion(
+    dandelion, transfer: bool = False, include_fields: Optional[Collection[str]] = None
```
grst, line 667: Now that the bug is fixed, I don't think we require any special casing.
To ensure `from_dandelion` and `read_airr` work in a consistent way, how about just passing `**kwargs` to `read_airr`. 

In that case, per default, `DEFAULT_AIRR_FIELDS` get imported from dandelion, if the user wants all, they can just pass include_fields=None.

scirpy/io/_datastructures.py
```diff
@@ -163,6 +163,8 @@ def add_serialized_chains(self, serialized_chains: str) -> None:
         if not _is_na2(serialized_chains):
             tmp_chains = json.loads(serialized_chains)
             for chain in tmp_chains:
+                # keep only keys present in _chain_fields
+                # chain = {k: v for k, v in chain.items() if k in self._chain_fields}
```
grst, line 167:
```suggestion
```

grst (commented):
Yes, exactely!

scirpy/io/_io.py
zktuong: sounds great. I'll sort that out.

scirpy/io/_datastructures.py
zktuong: this looks the same? is this just a black formatting thing? if so, i'll just make sure to format before my next push.

grst: Actually, I was suggesting to remove it entirely as it does nothing ;)

zktuong: ah of course. misread the colours.

scirpy/io/_io.py
zktuong: Hmm if i remove the special casing, it stops working. damn it.

grst:
The filtering for extra chains still doesn't work... I'll run it through with the debugger

image

grst:
I completely misread how https://github.com/icbi-lab/scirpy/blob/f25858842993b7ce039b00ccaef055ad78ee9746/scirpy/io/_datastructures.py#L240-L247 works during review. Actually, it wasn't filtering at all. 

I think it should be working now.

grst:
Does this work for you now?

zktuong:
Yes i think so! Let me do a bit of testing.

I also need to add a function to sanitize the transferred dataframe on dandelion's side.

grst:
ok, cool! 
I'll take a few days off next week, and will be back to review & merge on Thursday ;)

zktuong:
Looks like it's all working with `dandelion v0.1.4`! Although it took 2.5 hours to finish the tests on the macos3.8 test check.
Ready to merge when you're back.

grst:
> Although it took 2.5 hours to finish the tests on the macos3.8 test check

Let's hope that was a random hiccup :roll_eyes: 

Thanks again for the details bug report and your contribution! Will release v0.8 later.

PR #288, opened 2021-08-25
Improve infer_locus_from_gene_name

### LINKED ISSUE #285: TRAV/DV rearrangements
hey @grst, this was alerted to me:

_Since the TRD locus is embedded in the TRA locus, a few TRAV genes have been found rearranged, not only as expected to TRAJ but also to TRDD-TRDJ genes. The first five Homo sapiens TRAV genes found in that case received the designation TRAV/DV (this does not exclude that other TRAV genes may be found occasionally rearranged to TRDD-TRDJ). They include Homo sapiens TRAV14/DV4, TRAV29/DV5, TRAV23/DV6, TRAV36/DV7 and TRAV38-2/DV8 genes._(http://www.imgt.org/IMGTrepertoire/index.php?section=LocusGenes&repertoire=genetable&species=human&group=TRAV)

So currently, the infer locus code would not catch this:
https://github.com/icbi-lab/scirpy/blob/f0599abecc93c7db9ae1b0db268018d534ffe6da/scirpy/io/_io.py#L444-L472

These genes would be annotated with `TRA` locus, even after reassignment with igblastn. While not really an issue for TRA chains, it poses an issue for `TRD` chains because it seems `AirrCell` will drop the `d_call` from the affected contig, and wrongly assign it to `IR_VJ` columns rather than `IR_VDJ`:
image

I suppose something to catch it like this may work:
```python
queries = [q for q in [d_call, j_call, c_call] if pd.notnull(q)]
if re.search('TRAV.*/DV', v_call):
    if all([re.search('TRA', x) for x in queries])
        locus = 'TRA'
    elif all([re.search('TRD', x) for x in queries]) 
        locus = 'TRD'
```

grst:
Fix #285

### DIFF
--- a/scirpy/tests/test_io.py
+++ b/scirpy/tests/test_io.py
@@ -12,6 +12,7 @@ from scirpy.io import (
     upgrade_schema,
     AirrCell,
 )
+from scirpy.io._io import _infer_locus_from_gene_names
 from scirpy.io._util import _check_upgrade_schema
 from scirpy.util import _is_na, _is_false
 import numpy as np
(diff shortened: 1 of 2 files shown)


scirpy/tests/test_io.py
```diff
@@ -123,6 +124,60 @@ def test_airr_cell_empty():
...
+                "d_call": "TRDJ2*01",
+                "j_call": "nan",
+                "c_call": "TRDC",
+            },
+            "TRD",
+        ),
+        (
+            {
+                "v_call": "TRAV36/DV7*01",
+                "d_call": "TRDJ2*01",
```
zktuong, line 151: this is a `j_call`?

grst: good catch! fixed...

zktuong: should probably fill them with a dummy `d_call` as well. like `TRDD1*01`

scirpy/tests/test_io.py
```diff
@@ -139,17 +139,17 @@ def test_airr_cell_empty():
         (
             {
                 "v_call": "TRAV36/DV7*01",
-                "d_call": "TRDJ2*01",
-                "j_call": "nan",
+                "d_call": "nan",
```
zktuong, line 142: should probably fill them with a dummy d_call as well. like TRDD1*01

PR #316, opened 2021-12-28
fix: resolved dependency clash

BenTenmann:
numba requires numpy<=1.20

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ requires = [
     'anndata>=0.7.6',
     'scanpy>=1.6.0',
     'pandas>=1.0',
-    'numpy>=1.17.0',
+    'numpy>=1.17.0,<=1.20',
     'scipy',
     # parasail 1.2.1 fails to be installd on MacOS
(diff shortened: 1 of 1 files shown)


grst:
Hi @BenTenmann, 

thanks for the PR! Did you get an error during installation? 

I believe in principle this should not be necessary, as pip should figure out transitive dependencies by itself (i.e. scirpy requires numba, numba requires numpy < 1.20, so it won't install a more recent version of numpy)

BenTenmann:
Hi @grst 
No problem! Yes, in fact I did get an error when `pip` installing using a `python==3.8.2` environment, with a dependecy clash between `numba` and `numpy`.

grst:
Interesting... out of curiosity: Did you install scirpy into a clean environment, or was there maybe a more recent version of numpy already installed?

BenTenmann:
There was a more recent version of `numpy` (1.21.5) installed, hence the conflict.

grst:
Either way, guess this can't harm.

PR #317, opened 2022-01-08
fix/certificate

BenTenmann:
fixed an issue on macOS when downloading vdjdb data where ssl certificate was not available -- hence changed default certificate to unverified on macOS


grst:
I'm not sure if this is such a good idea, as this will disable certificate validation for the entire python session; and the certificate validation is there for [a reason](https://www.python.org/dev/peps/pep-0476/#id23). 

Also it doesn't seem to be a Mac problem *per se*, as it can be downloaded without issues on the MacOS CI.

PR #329, opened 2022-03-23
Create _io_bdrhapsody.py

Raphael-Gronauer:
io for BDrhapsody scRNA scTCR and scdCODE data, preprocessed with the BDrhapsody-pipeline

### DIFF
--- a/scirpy/io/_io_bdrhapsody.py
+++ b/scirpy/io/_io_bdrhapsody.py
@@ -1,96 +1,80 @@
 import pandas as pd
 import numpy as np
-import scirpy as ir 
+from ._datastructures import AirrCell
+from ._convert_anndata import from_airr_cells
+from ..pp._merge_adata import merge_with_ir
+
 import anndata
 import scanpy as sc
 
-class ImportBDrhapsodyData:
 
-    def load_tcr_seq(data):
+def read_bd_vdj_csv(data):
 
-        tcr_table = pd.read_csv(data, sep=",", index_col=None, na_values=["None"], true_values=["True"],comment='#')
-        tcr_table['productive_Alpha_Gamma'] = tcr_table.TCR_Alpha_Gamma_CDR3_Translation_Dominant.str.contains('\*', regex=True)
-        tcr_table['productive_Beta_Delta'] = tcr_table.TCR_Beta_Delta_CDR3_Translation_Dominant.str.contains('\*', regex=True)
-        print(tcr_table['productive_Beta_Delta'])
-        tcr_table
-        tcr_cells = []
-        for idx, row in tcr_table.iterrows():
-            cell = ir.io.AirrCell(cell_id=row["Cell_Index"])
-            alpha_chain = ir.io.AirrCell.empty_chain_dict()
-            beta_chain = ir.io.AirrCell.empty_chain_dict()
-            alpha_chain.update(
-                {
-                    "locus": "TRA",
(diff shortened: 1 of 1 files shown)


grst:
Closed and reopened because the CI was stuck...

grst:
Thanks for adding this! 

I'll give more detailed feedback later, but there are two more general things that need to be addressed: 
 * We need a test-case that checks the function works as expected. Could you add a small sample file (like the first hundred lines of a real file) to the `scirpy/tests/data/bd_rhapsody` folder? And a test function to https://github.com/scverse/scirpy/blob/master/scirpy/tests/test_io.py (just like the `test_read_10x_csv`). 
 * All code needs to be formatted using the [black](https://github.com/psf/black) formatter.

scirpy/io/_io_bdrhapsody.py
```diff
@@ -0,0 +1,96 @@
...
+        tcr_table = pd.read_csv(data, sep=",", index_col=None, na_values=["None"], true_values=["True"],comment='#')
+        tcr_table['productive_Alpha_Gamma'] = tcr_table.TCR_Alpha_Gamma_CDR3_Translation_Dominant.str.contains('\*', regex=True)
+        tcr_table['productive_Beta_Delta'] = tcr_table.TCR_Beta_Delta_CDR3_Translation_Dominant.str.contains('\*', regex=True)
+        print(tcr_table['productive_Beta_Delta'])
+        tcr_table
+        tcr_cells = []
+        for idx, row in tcr_table.iterrows():
+            cell = ir.io.AirrCell(cell_id=row["Cell_Index"])
+            alpha_chain = ir.io.AirrCell.empty_chain_dict()
+            beta_chain = ir.io.AirrCell.empty_chain_dict()
```
grst, line 20:
```suggestion
            cell = AirrCell(cell_id=row["Cell_Index"])
            alpha_chain = AirrCell.empty_chain_dict()
            beta_chain = AirrCell.empty_chain_dict()
```

scirpy/io/_io_bdrhapsody.py
```diff
@@ -0,0 +1,96 @@
...
+            alpha_chain.update(
+                {
+                    "locus": "TRA",
```
grst, line 23: I think we discussed this already, but there's no way to find out if this is a alpha or a gamma chain?

scirpy/io/_io_bdrhapsody.py
```diff
@@ -0,0 +1,96 @@
...
+        except:
+            adata = sc.AnnData(np.array(data_rna),var=genes)
+        
+        if tcr != 1:
+            data_tcr = ImportBDrhapsodyData.load_tcr_seq(data=tcr)
+            ir.pp.merge_with_ir(adata, data_tcr)
+        else:
+            pass
+
+        return adata
```
grst, line 96: The read_rna_seq_data should probably go to scanpy, as it is not TCR-specific. 

I'm not entirely sure about the decode, because it is somewhat TCR-related. But then again it is so trivial that it probably doesn't need a helper function anyway. 

I think I would just remove this part for now. Within [scverse](https://scverse.org/), we are thinking about a more general package for single-cell I/O. If we implement that package, I'll come back to this.

grst (changes requested):

scirpy/io/_io_bdrhapsody.py
Raphael-Gronauer: As far as i know from may data not

grst:
Closed in favor of #351

Issue #327, opened 2022-03-09
scverse datastructure for AIRR data


grst:
Now that scirpy is part of scverse, we could think of an improved data structure for scAIRR data. See also the discussion at https://github.com/theislab/scanpy/issues/1387. 

The challenge with scAIRR data is that
 * `1` cell can have `n` chains. Up to four of them are biologically meaningful but there could be more for technical reasons.
 * Each chain has a lot of fields. See the [AIRR rearrangement standard](https://docs.airr-community.org/en/latest/datarep/rearrangements.html). 

The current pragmatic solution is to store all fields in `adata.obs`. 
 * All columns from the airr rearrangement schema are repeated four times
 * Excess chains are serialized into JSON and stored in an extra column. These chains are not used by scirpy, but enable lossless conversions. 
 * The downside is that there can easily be 100+ columns in `adata.obs`. Also serializing excess chains is not really elegant. 
 * The advantage is that it works really well with scanpy, i.e. any AIRR variable can immediately be used for grouping, plotting etc. 

New options are
 * **mudata**. AIRR data could be saved as a separate modality. Even if we keep the current reprepsentation of a wide data frame, it would at least not clutter the rest of `adata.obs`. 
 * **awkward array support**. Allows storing an arbitrary number of values per row. See https://github.com/theislab/anndata/pull/647

The new representation should also aim at being a community standard for the scverse ecosystem and should build upon the AIRR rearrangement standard. Ideally, we could get additional stakeholders onboard, including conga, dandelion, tcrdist3 and possibly members of the AIRR community. 

- [x] what's the state of the AIRR single-cell schema? And what are its advantates over the rearrangement schema.

bcorrie:
Hi All, I am active in the AIRR Standards group, can try and answer questions. We are looking at integrating scirpy into the iReceptor Gateway (gateway.ireceptor.org) to do analysis on AIRR Cell data - so have a vested interest... 8-)

The AIRR Cell schema was just announce at the AIRR meeting this week, with a tagged (v1.4) release expected on github in the next few weeks. This is an "Experimental" schema, meaning that we think it is pretty stable, but will probably need some minor changes before we it is finalized, probably in an AIRR Standard v2.0 release sometime in the "near" future.

bcorrie:
[...]

PR #347, opened 2022-06-28
Update minimal dandelion version

### LINKED ISSUE #343: Cannot convert output from Scirpy to dandelion
### Description of the bug

I try to convert the AnnData after clonal assignment by Scirpy into dandelion format. The result showed that "field productive has invalid bool T + T". I would like to convert it for updating germline sequence of each BCR sequences using dandelion because I did not find this function in Scirpy. However, if you could suggest the other ways. Feel free to let me know.

### Minimal reproducible example

```python
import scirpy as ir

ABC_irdata_exclude_orphan_dandelion = ir.io.to_dandelion(ABC_irdata_exclude_orphan)
ABC_irdata_exclude_orphan_dandelion
```

### The error message produced by the code above

```pytb
~/.conda/envs/dandelion/lib/python3.8/site-packages/airr/schema.py in validate_row(self, row)
    276                 if spec == 'number':  self.to_float(row[f], validate=True)
    277             except ValidationError as e:
--> 278                 raise ValidationError('field %s has %s' %(f, e))
    279 
    280         return True

ValidationError: field productive has invalid bool T + T
```

### Version information

<details>
<summary>versions</summary>

```

```

</details>

grst:
Fix #343

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ test = [
     'pre-commit',
 ]
 dandelion = [
-    'sc-dandelion>=0.1.3',
+    'sc-dandelion>=0.2.3',
 ]
 diversity = [
     'scikit-bio>=0.5.7'
--- a/scirpy/tests/test_io.py
+++ b/scirpy/tests/test_io.py
@@ -294,8 +294,9 @@ def test_convert_dandelion(anndata_from_10x_sample):
     ddl = to_dandelion(anndata)
     anndata2 = from_dandelion(ddl)
 
-    ir_objs1 = to_airr_cells(anndata)
-    ir_objs2 = to_airr_cells(anndata2)
+    # dandelion reorders cell barcodes
+    ir_objs1 = sorted(to_airr_cells(anndata), key=lambda x: x.cell_id)
+    ir_objs2 = sorted(to_airr_cells(anndata2), key=lambda x: x.cell_id)
 
     assert len(ir_objs1) == len(ir_objs2) == anndata.shape[0]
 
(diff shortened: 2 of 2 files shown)


grst:
@zktuong, is there any chance the new version of dandelion reorders the cells during import or export? 

The round trip conversion fails now, but my first guess would be that just the order is shuffled.

zktuong:
ah, then yes. it will reorder the cells.

grst:
So this is intended behavior? May I ask why you are doing this? 

In any case no problem, I'll just update the test to ignore the order then.

zktuong:
it is. it's to make it more in line with how scirpy orders the contigs by umi counts for chains 1 and 2, and a way we are dealing with trimming/keeping non-productive contigs for some further analyses

grst:
> it is. it's to make it more in line with how scirpy orders the contigs by umi counts for chains 1 and 2, and a way we are dealing with trimming/keeping non-productive contigs for some further analyses

I'm confused. This sounds like reordering the chains again? I belive scirpy keeps the cells in the same order as the input file.

zktuong:
> I'm confused. This sounds like reordering the chains again? I belive scirpy keeps the cells in the same order as the input file.

oh? I was told by the people in the lab that used scirpy that it orders the chains based on umi counts, before selecting which is IR_*_1 and IR_*_2 and uses the highest umi value... so that's why i did it 😅

or does that only happen during clone definition step?

grst:
wait :grinning: 

it **does** order the **chains**. It does not reorder **cells/barcodes**.

zktuong:
ah ok. i think it should be correct overall - i just sort the entire airr contig table based on productive and duplicate_count, which may shift the cell order, but then scirpy reads/loads it in cell by cell and sorts the chain. 

that sounds ok to me.

grst:
Perfect, then I'll just update the test to ignore the order of cells. 

I hope we can converge on a mutual data structure eventually to get rid of conversion issues once and for all (#327). That will take some time unfortunately...

PR #360, opened 2022-08-18
Fix incompatibility with scipy 1.9

grst:


grst:
The fix should be good to go. The DoubleLookupNeighborFinder is self contained, and I checked that all functions previously working with the `nan` index now work with the integer index `-1`.

PR #365, opened 2022-10-03
Try fix CI

grst:


grst:
This https://github.com/igraph/python-igraph/issues/570

PR #374, opened 2022-12-19
Fix typos in IR query

Zethson:


grst (approved):
Good catch, thanks!

Zethson:
@grst 3.9 Windows has some issues. I restart the job already

grst:
Seems that didn't do it. 

Why is CI always such a pain in the neck :see_no_evil: I'll look into it!

Zethson:
Well, Windows CI annoyed me so much that I just killed it. Windows users should use WSL or, you know, go Linux. But this is not the issue for such a discussion, sorry :)

PR #377, opened 2023-01-12
Download IEDB and process it into an AnnData object

ausserh:
I added a function to download data from the IEDB and process it into an AnnData object. This can be used as an additional epitope database for the query function.

### DIFF
--- a/scirpy/datasets/__init__.py
+++ b/scirpy/datasets/__init__.py
@@ -206,6 +206,7 @@ def vdjdb(cached: bool = True, *, cache_path="data/vdjdb.h5ad") -> AnnData:
 
     return adata
 
+
 def iedb(cached: bool = True, *, cache_path="data/iedb.h5ad") -> AnnData:
     """\
     Download IEDBD and process it into an AnnData object.
@@ -226,19 +227,36 @@ def iedb(cached: bool = True, *, cache_path="data/iedb.h5ad") -> AnnData:
 
     with tempfile.TemporaryDirectory() as d:
         d = Path(d)
-        urllib.request.urlretrieve("https://www.iedb.org/downloader.php?file_name=doc/receptor_full_v3.zip", d / "receptor_full_v3.zip")
+        urllib.request.urlretrieve(
+            "https://www.iedb.org/downloader.php?file_name=doc/receptor_full_v3.zip",
+            d / "receptor_full_v3.zip",
+        )
         with zipfile.ZipFile(d / "receptor_full_v3.zip") as zf:
             zf.extractall(d)
-        tcr_table = pd.read_csv(d / "receptor_full_v3.csv", index_col=None, sep=",", na_values=["None"], true_values=["True"],)
+        tcr_table = pd.read_csv(
+            d / "receptor_full_v3.csv",
+            index_col=None,
+            sep=",",
+            na_values=["None"],
+            true_values=["True"],
+        )
 
-    tcr_table.loc[tcr_table["Chain 1 CDR3 Curated"].isna(),'Chain 1 CDR3 Curated'] = tcr_table["Chain 1 CDR3 Calculated"]
+    tcr_table.loc[
+        tcr_table["Chain 1 CDR3 Curated"].isna(), "Chain 1 CDR3 Curated"
+    ] = tcr_table["Chain 1 CDR3 Calculated"]
 
-    tcr_table.loc[tcr_table["Chain 2 CDR3 Curated"].isna(),'Chain 2 CDR3 Curated'] = tcr_table["Chain 2 CDR3 Calculated"]
+    tcr_table.loc[
+        tcr_table["Chain 2 CDR3 Curated"].isna(), "Chain 2 CDR3 Curated"
+    ] = tcr_table["Chain 2 CDR3 Calculated"]
 
     tcr_table_T = tcr_table[(tcr_table["Response Type"] == "T cell")]
 
-    tcr_table_T["Chain 1 CDR3 Curated"] = tcr_table_T["Chain 1 CDR3 Curated"].str.upper()
-    tcr_table_T["Chain 2 CDR3 Curated"] = tcr_table_T["Chain 2 CDR3 Curated"].str.upper()
+    tcr_table_T["Chain 1 CDR3 Curated"] = tcr_table_T[
+        "Chain 1 CDR3 Curated"
+    ].str.upper()
+    tcr_table_T["Chain 2 CDR3 Curated"] = tcr_table_T[
+        "Chain 2 CDR3 Curated"
+    ].str.upper()
 
     tcr_table = tcr_table_T
 
--- a/scirpy/datasets/__init__.py
+++ b/scirpy/datasets/__init__.py
@@ -255,16 +255,20 @@ def iedb(cached: bool = True, *, cache_path="data/iedb.h5ad") -> AnnData:
         ]
     )
 
-    iedb_df.loc[
-        iedb_df["Chain 1 CDR3 Curated"].isna(), "Chain 1 CDR3 Curated"
-    ] = iedb_df["Chain 1 CDR3 Calculated"]
+    # If no curated CDR3 sequence or V/D/J gene is available, the calculated one is used.
+    def replace_curated(input_1, input_2):
+        Calculated_1 = input_1.replace("Curated", "Calculated")
+        tcr_table.loc[tcr_table[input_1].isna(), input_1] = tcr_table[Calculated_1]
+        tcr_table[input_1] = tcr_table[input_1].str.upper()
+        Calculated_2 = input_2.replace("Curated", "Calculated")
+        tcr_table.loc[tcr_table[input_2].isna(), input_2] = tcr_table[Calculated_2]
+        tcr_table[input_2] = tcr_table[input_2].str.upper()
+        return tcr_table
 
-    iedb_df.loc[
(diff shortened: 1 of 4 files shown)


grst:
Hi Markus, 

thanks a lot for this PR and for your message! 
The current problem is indeed just formatting. You can fix it with the [black](https://github.com/psf/black) formatter. There's also some instructions in the [scirpy developer docs](https://scverse.org/scirpy/latest/developer_docs.html#formatting-code). 

Let me know if you are stuck. 

Best, 
Gregor

scirpy/datasets/__init__.py
```diff
@@ -205,3 +205,110 @@ def vdjdb(cached: bool = True, *, cache_path="data/vdjdb.h5ad") -> AnnData:
...
+        beta_chain = AirrCell.empty_chain_dict()
+        alpha_chain.update(
+            {
+                "locus": "TRA",
+                "junction_aa": row["Chain 1 CDR3 Curated"],
+                "junction": "None",
+                "consensus_count": "None",
+                "v_call": row["Curated Chain 1 V Gene"],
+                "j_call": row["Curated Chain 1 J Gene"],
+                "productive": "True",
```
grst, line 276: Same here, please use an actual boolean instead of the string `"True"`.

grst (changes requested):
Here's the first round of review with some more detailed comments. 

General things:
 * for a new function to show up in the documentation, you'll need to add an entry in [`api.rst`](https://github.com/scverse/scirpy/blob/master/docs/api.rst)
 * Would it be possible to also include BCR sequences? 
 * What does it mean if there are multiple rows in the IEDB table that have the same receptor ID? Should they be represented by multiple entries in the `adata` object?

grst:
btw, don't worry about that failing `conda` test. I'm not sure yet what's the problem, but it is not your fault.

ausserh:
Hi Gregor,
thanks for your suggestions, I updated my request accordingly.
Regarding your two questions, I will have a closer look and then come back to you.
Best,
Markus

ausserh:
Ah, the formatting again, will fix it!

ausserh:
Good morning Gregor,
regarding your questions:

Around 2.6 % of all entries are BCRs, I do not filter them out when processing the data. When performing the actual query, I would assume that we do not need to worry that TCR CDR3 sequences from the query wrongly match with the BCR ones in the dataset. Within the dataset, none of the TCR and BCR CDR3 sequences match.

8 % of entries have a receptor ID that is present more than once, however most of these receptors still differ in other aspects. From these 8%, around 25% have the same chain1 CDR3, chain2 CDR3, antigen organism, antigen and cell type, we could think about removing these, because the rest would then have the same receptor ID, but would still differ in one of these essential aspects.
From all the entries, around 1% are completely duplicated, I can remove them right at the beginning.

scirpy/datasets/__init__.py
```diff
@@ -205,3 +205,106 @@ def vdjdb(cached: bool = True, *, cache_path="data/vdjdb.h5ad") -> AnnData:
...
+    logging.info("Converting to AnnData object")
+    iedb = from_airr_cells(tcr_cells)
+    tcr_table = tcr_table.set_index(iedb.obs.index)
+    iedb.obs["Antigen"] = tcr_table["Antigen"]
+    iedb.obs["Organism"] = tcr_table["Organism"]
+    iedb.obs["Chain 1 Type"] = tcr_table["Chain 1 Type"]
+    iedb.obs["Chain 2 Type"] = tcr_table["Chain 2 Type"]
+    iedb.obs["Response Type"] = tcr_table["Response Type"]
+    iedb.obs["Reference IRI"] = tcr_table["Reference IRI"]
+    iedb.obs["Epitope IRI"] = tcr_table["Epitope IRI"]
```
grst, line 302: you could assign these directly in the previous `for` loop to the `cell` object: 
```python
cell = AirrCell(cell_id=row["Receptor ID"])
cell["Antigen"] = row["Antigen"]
```

that's a bit easier to follow and eliminates an additional error source.

grst (commented):
> Around 2.6 % of all entries are BCRs, I do not filter them out when processing the data.

I think you do by subsetting to `"Response Type" == "T cell"`. IMO it would be better to keep them and set the `locus` based on `Chain1/2 type` accordingly. Also to distinguish alpha/beta vs. gamma/delta TCR. Here's [the list of supported locus types](https://scverse.org/scirpy/latest/glossary.html#term-Chain-locus). 

> 8 % of entries have a receptor ID that is present more than once, however most of these receptors still differ in other aspects. From these 8%, around 25% have the same chain1 CDR3, chain2 CDR3, antigen organism, antigen and cell type, we could think about removing these, because the rest would then have the same receptor ID, but would still differ in one of these essential aspects.

ok, I think it makes sense to only keep unique entries based on the columns we include in the final object. 
Another important point here is that in this case `receptor_id` shouldn't be used as `cell_id`. Instead, you could e.g. use the row index of the pandas data frame. Every unique receptor needs to be stored as it's own "cell", otherwise attributes like `Antigen` get lost (only the last one will be preserved I guess). If there are more than two chains they get moved into the `extra_chains` column (scirpy only support up to two chains per locus per cell). 

image

grst:
Ok, getting there. I found another issue, but already fixed it myself: IEDB does only specify "light" chains for BCRs, but scirpy needs to know if they are IGL (lambda) or IGK (kappa) loci. To fix this, I re-used a function to infer the locus from the gene names that I already used in other IO modules. 

I also renamed the `tcr_table` to `iedb_df`, because it does not only contain TCRs. 

Here's the `chain_qc` report for `iedb` now. Does this look reasonable? 
| pairing | receptor type |
|-------|-----|
|image|image|

scirpy/datasets/__init__.py
```diff
@@ -205,3 +206,138 @@ def vdjdb(cached: bool = True, *, cache_path="data/vdjdb.h5ad") -> AnnData:
...
+    iedb_df.loc[
+        iedb_df["Chain 1 CDR3 Curated"].isna(), "Chain 1 CDR3 Curated"
+    ] = iedb_df["Chain 1 CDR3 Calculated"]
+
+    iedb_df.loc[
+        iedb_df["Chain 2 CDR3 Curated"].isna(), "Chain 2 CDR3 Curated"
+    ] = iedb_df["Chain 2 CDR3 Calculated"]
+
+    iedb_df["Chain 1 CDR3 Curated"] = iedb_df["Chain 1 CDR3 Curated"].str.upper()
+    iedb_df["Chain 2 CDR3 Curated"] = iedb_df["Chain 2 CDR3 Curated"].str.upper()
```
grst, line 267: One more request: Do you think it would make sense to also merge the Calculated/Computed for the V/D/J gene names?

ausserh: Thank you for fixing the light chain issue!
Yes, makes sense also to replace the V/D/J gene names if not present with the calculated ones, this is included in my new commit, let me know what you think.

scirpy/datasets/__init__.py
```diff
@@ -255,16 +255,20 @@ def iedb(cached: bool = True, *, cache_path="data/iedb.h5ad") -> AnnData:
...
+        tcr_table[input_1] = tcr_table[input_1].str.upper()
+        Calculated_2 = input_2.replace("Curated", "Calculated")
+        tcr_table.loc[tcr_table[input_2].isna(), input_2] = tcr_table[Calculated_2]
+        tcr_table[input_2] = tcr_table[input_2].str.upper()
+        return tcr_table
+
+    tcr_table = replace_curated("Chain 1 CDR3 Curated", "Chain 2 CDR3 Curated")
+    tcr_table = replace_curated("Curated Chain 1 V Gene", "Curated Chain 2 V Gene")
+    tcr_table = replace_curated("Curated Chain 1 D Gene", "Curated Chain 2 D Gene")
+    tcr_table = replace_curated("Curated Chain 1 J Gene", "Curated Chain 2 J Gene")
```
grst, line 271: Good idea to make a function! 

One point here is that you modify the object `tcr_table` in the function directly.
This is perfectly fine in this case, but you don't need to return and reassign the table. 

codestyle nitpick: `Calculated_` --> `calculated_` as it represents a variable, not a class.

ausserh: Thank you for your suggestions, but I am not sure if I understand correctly. If I don't use return, the function doesn't return anything, if I don't assign the output to a variable, it is not stored. What should I change?

Okay I will change the variable name, also, I need to change "tcr_table" to "iedb_df", "tcr_table" doesn't exist anymore since you changed the variable name earlier.

grst: There's an important difference between primitive types (int, etc.) and objects (e.g. dict, DataFrame, ...).
An object is always passed as a reference, that means you can change the content of the original object. 

Check out the following example: 

```python
a = 5
b = {"b": 5}

def do_stuff():
    a = 7
    b["b"] = 7
    
do_stuff()
print(a)
print(b)
```

Maybe also check out [this section](https://fabienmaussion.info/scientific_programming/week_03/01-Import-Scopes.html) of the python course at uni innsbruck. Maybe you have even a chance taking it, the online version looks excellent.

ausserh: Okay, this example is very clear, I understand. I updated the code. Thanks for the Link, the course seems indeed well-made.

grst (approved):
LGTM now, thanks a lot for your work!

ausserh:
Great! Thank you for your help!

grst:
We are live! 
```
pip install scirpy==0.12
```

PR #396, opened 2023-04-13
Switch back to markdown issue templates

grst:
I have the feeling the bug report form is too intimidating, and I rather receive an incomplete bug report than none at all.


Zethson:
I'm surprised. I really really like the new templates (also as an issue submitter!)

grst:
I have the feeling that I receive less bug reports since I switched to the yaml template... But this may (1) be subjective and (2) confounded.

PR #436, opened 2023-07-21
dependencies: change python-igraph to igraph

szhorvat:
Please see https://github.com/igraph/python-igraph/issues/699 for the explanation.


grst:
Thanks for raising this! 

The conda-forge package would still by `python-igraph` though?

szhorvat:
> The conda-forge package would still by `python-igraph` though?

Yes, that one stays `python-igraph`, see https://github.com/conda-forge/python-igraph-feedstock/issues/42 for the reasoning. I hope this is not causing any problems for you.

grst:
No problems, I just wanted to know if I had to update the conda recipes as well :)

PR #456, opened 2023-10-18
Add fastalignment metric

### LINKED ISSUE #304: Speed up ir_dist
Parasail is highly optimized - but it is invoked for every sequence individually from Python which creates a bottleneck. 
We could solve this issue by integrating with external tools that have been developed for scalable comparison of immune-cell repertoires (both for levenshtein and alignment distance). 

Possible methods to consider
 - [ ] MMSeqs2: Originally designed for metagenomics, this is a very fast multiple sequence alignment tool. See some of the discussion below. 
 - [ ] compAIRR ( https://www.biorxiv.org/content/10.1101/2021.10.30.466600v2.full)
 - [ ] clusTCR (https://svalkiers.github.io/clusTCR/)
 - [ ] tcrdist3
 - *...* (there's probably many more)

Ideally, we could provide wrappers for several of them.

tobgan:
Closes #304

-   [x] CHANGELOG.md updated
-   [x] Tests added (For bug fixes or new features)


src/scirpy/ir_dist/metrics.py
```diff
@@ -478,3 +478,200 @@ def _self_alignment_scores(self, seqs: Sequence) -> dict:
...
+        *Proline* (`P`) already results in a distance of `15`.
+
+        We are still lacking empirical data up to which distance a CDR3 sequence still
+        is likely to recognize the same antigen, but reasonable cutoffs are `<15`.
+
+    Choosing an expected penalty:
+        The choice of an expected penalty is likely influenced by similar considerations as the
+        other parameters. Essentially, this can be thought of as a superficial (dis)similarity
+        measure. A higher value more strongly penalizes mismatching characters and is more in line
+        with looking for closely related sequence paris, while a lower value is more forgiving
```
grst, line 522:
```suggestion
        with looking for closely related sequence pairs, while a lower value is more forgiving
```

src/scirpy/ir_dist/metrics.py
```diff
@@ -478,3 +478,200 @@ def _self_alignment_scores(self, seqs: Sequence) -> dict:
...
+            "pam190": 2.0,
+            "pam200": 2.0,
+        }
+
+        self.estimated_penalty = (
+            estimated_penalty
+            if estimated_penalty is not None
+            else penalty_dict[subst_mat]
+            if subst_mat in penalty_dict.keys()
+            else 0.0
```
grst, line 601: can be simplified to

```suggestion
            else penalty_dict.get(subst_mat, 0.0)
```

I would even consider raising an error if the substitution matrix is unnown and no penalty is specified.

src/scirpy/ir_dist/metrics.py
```diff
@@ -478,3 +478,200 @@ def _self_alignment_scores(self, seqs: Sequence) -> dict:
...
+        estimated_penalty: float = None,
+    ):
+        if cutoff is None:
+            cutoff = 10
+        super().__init__(cutoff, n_jobs=n_jobs, block_size=block_size)
+        self.subst_mat = subst_mat
+        self.gap_open = gap_open
+        self.gap_extend = gap_extend
+
+        penalty_dict = {
```
grst, line 559: Could you please add a comment how you came up with these values? Maybe even in the docstring that the users can see it.

grst (changes requested):
Hi @tobgan, 

thank you so much for adding this! 
The code you added looks good, but there are a few point that still need some work to make this more easily accessible to users. I'm happy to guide you, but I can also work on those tasks myself if you prefer. 

 - [ ] documentation: The new class should be added to the API docs
 - [ ] tutorial: the new method should be mentioned in the docs and probably even be suggested as the standard method. 
 - [ ] The code needs unit tests (it's mostly applying the existing tests to the new method as well and adjusting the expected values)

Before we tackle those I have a more general question: Do I remember correctly, that if *only* the length filtering is applied there are no losses and the results are *exactly* the same as with the plain `AlignmentDistanceCalculator` but faster? (obviously not as fast as the full method). Because in that case I would consider getting rid of the old AlignmentDistanceCalculator altogether and just use the new class with different parameters for `alignment` and `fastalignment`. What do you think?

Best, 
Gregor

grst:
Hi @tobgan -- would be great if I could get your feedback on the points raised above.

tobgan:
Hi @grst,

I'm very sorry for the belated response, I have been dealing with some health issues lately.

Regarding the length filter - yes, the results are exactly the same, at least in practice. _Theoretically_, it could still result in some loss, e.g. with unusually high-scoring mismatches (as, for example, in PAM500, where the mismatch N-D scores higher than the match N-N), but this has not happened in any of my test runs and I would wager to say it is unlikely to happen with real data, and even less likely to do so to a relevant degree. So I think replacing the `AlignmentDistanceCalculator` altogether would make sense.

If you agree, I would then just make the `FastAlignmentDistanceCalculator` the new `AlignmentDistanceCalculator`. I will also add the unit tests, but I could use some pointers on how to best add to the documentation and tutorial.

Best,
Tobias

src/scirpy/ir_dist/metrics.py
tobgan: Agreed, I think raising an error would be better

grst:
Thanks, that sounds good. Honestly, I doubt anyone has ever changed the substitution matrix to anything other than blosum62, so that should be fine. 

I would then suggest that in `pp.ir_dist`
 * specifying `metric="alignment"` sets the parameters of the `FastAlignmentDistanceCalculator` such that the result is identical to the previous `AlignmentDistanceCalculator` for backwards compatibility (ignoring the PAM500 caveat)
 * specifying `metric="fastalignment"` uses the same class, but sets the parameters such that the heuristic is used for additional speedup. 

Regarding the API docs, the list of functions is here: https://github.com/scverse/scirpy/blob/main/docs/api.rst?plain=1#L298

For the tutorial, it's probably easiest if I take care of that myself as a final step.

grst:
Perfect, thanks! I'll go over the documentation one more time myself and update the tutorial and then merge this!

grst:
Adding the tables with the expected penalty here for future reference: 

**Wu2020 3k dataset**
image

**Wu2020 full dataset**
"using the nighest expected penalty which kept loss <1% for the 3k dataset"
image

grst (approved):
Wrapped up the open things... Looks good to me now! 

Some CI job is failing, but it doesn't look related to this PR. 

Thanks again @tobgan!

PR #470, opened 2024-01-07
Speed up clonotype distance calculation

### LINKED ISSUE #368: speed up define_clonotypes
### Description of feature

The define_clonotypes function scales badly. There are two problems with it
  * it could be faster (while it relies heavily on numpy, there are parts implemented in Python)
  * parallelization doesn't work properly with large data. Due to how multiprocessing is implemented in Python, parallelization involves a lot of copying. If parallelization worked properly, the speed would still be bearable if one throws enough cores at the problem. 

### Where's the bottleneck of the function? 

INPUT: 
 * 2 distance matrices, one for unique VJ sequences, one for unique VDJ sequences

OUTPUT: 
 * a clonotype id for each cell

CURRENT IMPLEMENTATION: 
 1) compute unique receptor configurations (i.e. combining cells with the same sequences into a single entry) (fast)
 2) build a lookup table from which the neighbors of each cell can be retrieved (fast enough)
 3) loop through all unique receptor configurations and find neighbors (SLOW)
 4) build a distance matrix (fast)
 5) graph partition using igraph (fast)

ALTERNATIVE IMPLEMENTATIONS I considered but discarded
 * reindexing sequence distance matrices such that they match the table of unique receptor configurations
 * Then perform matrix operations to combine primary/secondary and TRA/TRB matrices. 
 * The problem with this  approach is that large dense blocks in the sparse matrices can arise if many unique receptors have the same sequence (e.g. same TRA but different TRB). 

### Possible solutions

 * fix parallelization ([shared memory](https://docs.python.org/3/library/multiprocessing.shared_memory.html))
 * reimplement using jax/numba (this may also solve the parallelization and provide GPU support)
 * Combine 2-4 into a single step (maybe possible with sequence embedding -- see #369 ). Note that this would be an alternative route and wouldn't replace ir_dist/define_clonotypes completely. 
 * Special-casing: In the case of omniscope data (which only has TRB chains), the problem simplifies to reindexing a sparse matrix. If using only one pair of sequences per cell, the problem is likely also simpler.

### LINKED ISSUE #507: clone definition purely using CDR3 sequence
Hi @grst,

I've been asked recently to create clonotype definitions purely using the identical/similar CDR3 sequence (and not considering any V/J gene information) and wanted to check with you whether this could be something useful to implement here as well? As far as i can tell, there's currently only `same_v_gene` but i guess we can add another `same_j_gene`?

felixpetschko:
Close #368 
Close #507 

compute_distances new version added


grst:
Hi @felixpetschko, 

I just had a look at the failing tests. Most of them are indeed direct or indirect failures due to the new distance calculator. Some of them are testing the API of the `DoubleLookupNeighborFinder` and are expected to fail because you changed the API (for good reasons). Those we'll have to adapt or remove eventually. 

At first, I suggest you take a look at the tests in `test_clonotypes` -- it seems they cover some edge cases where your implementation throws an error. 

image

LMK if anything is unclear! 

Gregor

felixpetschko:
@grst Now I could make all tests pass without changing them - I didn't break the API. There are still some documentation and other refinements needed.
The next step would be to implement j_gene matching which I already did for a version that @MKanetscheider uses for his analysis by just copying the v_gene related parts of the code. Should I also do it like that here or should we introduce a more general concept, eg by just passing the column suffixes (eg ["v_call", "j_call",...]) instead of the booleans (same_v_gene, same_j_gene). I don't know if it's biologically possible to have more than those 2.

grst:
Hi @felixpetschko, 

Kelvin suggests that only v and j calls are relevant, and I'd follow that advice and go with 
```
same_v_gene: bool = True,
same_j_gene: bool = True
```
as an interface.

I'll leave it to you if you think it's better solved with code duplication or generalization under the hood.

felixpetschko:
I implemented j_gene matching now with that interface. test cases (for j_gene matching) and documentation are still needed.
Is there still something else missing for the clonotype distance calculation you are aware of @grst ?

src/scirpy/ir_dist/_clonotype_neighbors.py
```diff
@@ -203,125 +231,205 @@ def _make_chain_count(self, clonotype_table) -> dict[str, int]:
...
+        for value in lookup.values():
+            has_distance_table += value
+
+        has_distance_mask = has_distance_table
+        has_distance_mask.data = np.ones_like(has_distance_mask.data)
+
+        def csr_min(a, b):
+            max_value_a = np.max(a.data, initial=0)
+            max_value_b = np.max(b.data, initial=0)
+            max_value = np.int16(np.max([max_value_a, max_value_b]) + 1)
```
grst, line 267: Why the `np.int16` here?

grst:
Yeah, a couple of docstrings seem to got lost or need updating, but I can't think of anything else right now.

src/scirpy/ir_dist/_clonotype_neighbors.py
felixpetschko:
```
        def csr_min(a, b):
            max_value_a = np.max(a.data, initial=0)
            max_value_b = np.max(b.data, initial=0)
            max_value = np.int16(np.max([max_value_a, max_value_b]) + 1)
            min_mat_a = sp.csr_matrix((a.data.astype(np.int16), a.indices, a.indptr), shape=a.shape)
            min_mat_a.data -= max_value
            min_mat_b = sp.csr_matrix((b.data.astype(np.int16), b.indices, b.indptr), shape=b.shape)
            min_mat_b.data -= max_value
            a_smaller_b = min_mat_a < min_mat_b
            return b + (a - b).multiply(a_smaller_b)
```

The idea behind this function is to get the minimum values of 2 csr matrices while ignoring 0 values with built-in functions. The trick here is to shift the data arrays by the maximum value, such that you get negative values. Then you can use the built in function "<" and ignore the 0 values. int16 is needed to allow negative values and cover at least all values that are are possible with uint8.

grst: Ok, thanks for explaining! Would you mind adding this as a comment in the code? 

Should the distance matrices not be of uint8 type, would this fail, or lead to incorrect results? If it doesn't fail explicitly, I'd add an assertion for the data type.

felixpetschko: > Ok, thanks for explaining! Would you mind adding this as a comment in the code?

Yes, I will add more documentation to all of the functions!

> Should the distance matrices not be of uint8 type, would this fail, or lead to incorrect results? If it doesn't fail explicitly, I'd add an assertion for the data type.

I will have to take a look at the datatypes in general. Probably i should already restrict the data type to uint8 at the distance matrix computation. Do you think anyone would want to set a cutoff > 255 (uint max)? That's probably too high even for the tcrdist.

grst: > Probably i should already restrict the data type to uint8 at the distance matrix computation

yes, but it wouldn't prevent anyone from building their own distance calculator class or just overwriting the distance values in `AnnData.uns` for whatever reason. So we still need checks here. 

> Do you think anyone would want to set a cutoff > 255 (uint max)? That's probably too high even for the tcrdist.

I don't think this would make sense for any of the metrics we use.

grst:
I dropped support for python 3.9 in the main branch which changed some of the formatting rules (most notably `Union[A, B]` is now simply `A | B`. 

I resolved the (few) merge conflicts and hopefully didn't break anything.

felixpetschko:
Hi @grst , now I added some documentation and did some variable renaming for clarification. I also added testing for the j_gene matching.

```
    if partitions == "leiden":
        part = g.community_leiden(
            objective_function="modularity",
            resolution_parameter=resolution,
            n_iterations=n_iterations,
        )
    elif partitions == "fastgreedy": # <--
        part = g.community_fastgreedy().as_clustering()
    else:
        part = g.clusters(mode="weak")
```
I also allowed to use the "fastgreedy" algorithm for the graph partitioning since it's used for the BCR analysis. However I have been told that choosing a different partitioning method doesn't make a big difference in most cases.

There still seems to be a test case that has some problems with the python versions (conda / test (ubuntu-latest, 3.12) (pull_request)) Could you maybe take a look at it?

Besides from that I am done with my implementations so far. If everything is fine for you I would make some final test runs on the cluster tomorrow. Then we can finally finish this PR :)

grst (approved):
Thanks, LGTM now! 

I'll merge it and make a release on Monday just in time for the conference. 

The conda test failure is not your fault. I just retriggered it, if it doesn't fix itself I'll look into it or we can ignore it for now.

PR #476, opened 2024-01-09
Calculate junction_aa in IEDB

### LINKED ISSUE #469: IEDB database cdr3_aa stored as junction_aa
**Describe the bug**
IEDB database provides cdr3_aa sequences instead of junction_aa sequences. scirpy database import code puts this in the slot for junction_aa. The sequence would then be missing the flanking amino acids. This makes it inaccurate to do identity string matching with ir_dist/ir_query.

**To Reproduce**
https://github.com/scverse/scirpy/blob/d862cf35740a79e91f95c49ce6da1fbe280f8c1c/src/scirpy/datasets/__init__.py#L338C1-L371C10

**Expected behaviour**
Is there a way to format cdr3 into junction sequence? like add the missing "C" at the beginning based on V call? Do you have advice on how to do this?

grst:
Closes #469

-   [x] CHANGELOG.md updated


grst:
This needs another solution as the protein sequence is missing too often.

PR #481, opened 2024-01-16
Speed up hamming distance calculation and add normalized hamming distance

### LINKED ISSUE #256: Normalized Hamming Distance
It would be better to have a normalization options in distance metric for BCR support. Still not clear on how to use the abstract class DistanceCalculator (tutorial could help)

felixpetschko:
Close #256 

Faster implementation of hamming distance calculation added to HammingDistanceCalculator. Boolean parameter normalize added to HammingDistanceCalculator constructor. If normalize is set to true, the hamming distance values are divided by the sequence length. New metric normalized_hamming added, which uses the HammingDistanceCalculator with normalize set to true.


grst (changes requested):
Hi @felixpetschko, 

could you please
 * fix the merge conflicts
 * merge in the latest changes from the `main` branch
 * add a testcase for the normalized hamming distance somewhere around here: https://github.com/scverse/scirpy/blob/f2531118a93dc9d36f137c098da7d43c73dfad7f/src/scirpy/tests/test_ir_dist_metrics.py#L209

grst:
Superseded by #512

PR #511, opened 2024-04-25
TCRdist performance improvements

felixpetschko:
TCRdist performance improvements added and joblib used


src/scirpy/ir_dist/metrics.py
```diff
@@ -677,16 +707,29 @@ def calc_dist_mat(self, seqs: Sequence[str], seqs2: Optional[Sequence[str]] = No
         if seqs2 is None:
             seqs2 = seqs
 
+        seqs = np.array(seqs)
+        seqs2 = np.array(seqs2)
+        is_symmetric = np.array_equal(seqs, seqs2)
+        n_blocks = self.n_jobs * 2
```
grst, line 713: How bad would it be for performance to increase the number of blocks? Main (only?) advantage would be that the user can see a meaningful progressbar as the blocks are being processed.

grst:
Looks great, thanks! Just one question above.

src/scirpy/ir_dist/metrics.py
felixpetschko: I tried to have 20 blocks per job and the execution time doubled in some cases. One of the reasons for that could be that the numba function has to be compiled again for each block which takes a few seconds each time. I set the number of blocks per job to 2 to compensate a little bit for the uneven workload of the row processing in the case of a triangular matrix. That way most of the time each job will get a long and a correspondingly short row. I think with this implementation it wouldn't be a good idea to increase the number of blocks :/

grst: ok, thanks for checking. Let's keep it as is then!

Issue #450, opened 2023-08-18
Can't install scirpy on Apple Silicon


a-munoz-rojas:
**Describe the bug**
Unable to install scirpy on Apple silicon using conda/mamba

**To Reproduce**
`mamba install scirpy --channel conda-forge --channel bioconda`

**Expected behaviour**
Normal installation

**System**

-   OS: MacOSX (Apple Silicon Arm64)
-   Python version 3.9
-   Versions of libraries involved scirpy 0.13

**Additional context**
The installation fails because of a dependency on parasail-python, which doesn't seem to be enabled in for osx-arm64 in conda? This is the error message I get:
```
bioconda/osx-arm64                                 129.0 B @   1.2kB/s  0.1s
bioconda/noarch                                      4.6MB @   4.6MB/s  1.1s
conda-forge/osx-arm64                                7.2MB @   4.4MB/s  1.8s
conda-forge/noarch                                  13.7MB @   4.6MB/s  3.3s

Pinned packages:
  - python 3.9.*


Could not solve for environment specs
The following package could not be installed
└─ scirpy   is uninstallable because it requires
   └─ parasail-python  , which does not exist (perhaps a missing channel).
```

grst:
I'm afraid bioconda does currently not compile packages for ARM64. See also these issues:

https://github.com/bioconda/bioconda-utils/issues/706
https://github.com/bioconda/bioconda-recipes/issues/23454

You should still be able to install scirpy using `pip install`.

grst:
Thanks for the heads-up! Creating an issue in parasail sounds good! If you do, please tag me that I'm kept in the loop!

djlisko01:
Hello, I had the same issue on my Apple M2 Pro and fixed it by running the following code before installing scirpy:
` brew install automake libtool autoconf`

FvdBre:
I am quite new to python, so my apologies if I miss the obvious ;-) . 
I have tried this approach as well using conda to install the dependencies. 

1. `conda install automake libtool autoconf`
2. `pip install scirpy`

However, I get the same error:

`  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for parasail
  Running setup.py clean for parasail
Failed to build parasail
ERROR: Could not build wheels for parasail, which is required to install pyproject.toml-based projects`

Even when I manually download the parasail source code, compiled it, and then installed the Python bindings.

Versions:

- m4                        1.4.18               h1230e6a_1 
[...]

PR #534, opened 2024-08-09
Logoplots

### LINKED ISSUE #12: Sequence Logos
In GitLab by @grst on Jan 24, 2020, 14:06

Here's a quite recent re-implementation of sequence logos in Python that looks promising:
https://github.com/jbkinney/logomaker

We will also require an algorithm for multiple sequence alignment in addition to the pairwise one that we have already. 

Things to discuss: 
- [ ] What sequences to use (TCRA/B primary only?) 
- [ ] What happens if a cell lacks the alpha/beta chain?

MKanetscheider:
Closes #12 
Added the file for sequence motif analysis via logoplots. It works as a wrapper function to the palmotif package. Also added palmotif now to the dependencies together with IPython. The latter was used to help with SVG visualization
-   [x] CHANGELOG.md updated
-   [x] Tests added (For bug fixes or new features)
-   [x] Tutorial updated (if necessary)
-   [x] also update TCR tutorial?


grst:
The conda tests are failing because the dependency for the conda tests needs to be declared separately here:
https://github.com/scverse/scirpy/blob/6160f375d59ca299577040904141ca0b0ee8a832/.conda/meta.yaml#L23

If you add palmotif in that list it shoud be ok.

pyproject.toml
```diff
@@ -42,6 +42,8 @@ dependencies = [
     'pooch>=1.7.0',
     'pycairo>=1.20; sys_platform == "win32"',
     'joblib>=1.3.1',
+    'palmotif',
+    'IPython',
```
grst, line 46: Why is the IPython dependency necessary?

MKanetscheider: Palmotif returns a SVG object, which is by default saved as such accordingly. What I did here was that I used the IPython.display's function SVG to directly display the SVG in my notebook and not always have to save it as a file. This was a workaround that I have almost forgotten, and I am not sure that this is even necessary...although I like to be able to directly investigate plots inside the notebook. However, I noticed that this is probably not the smartest way to do as I do not offer the possibility to save the SVG to a file any more and I might adapt this accordingly. Either way, I am not sure if SVG is a handy format for users to deal with... do you know if there is a possibility to save it as a png file or any other format that might be more convenient to deal with?

MKanetscheider: Will have a look into this, but sounds promising :D

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,177 @@
...
+import palmotif as palm
+import pandas as pd
+from IPython.display import SVG
+
+from scirpy.get import airr as get_airr
+from scirpy.util import DataHandler
+
+
+@DataHandler.inject_param_docs()
+def logoplot_cdr3_motif(
```
grst, line 13: please add the function to the API documentation: https://github.com/scverse/scirpy/blob/main/docs/api.rst

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,177 @@
...
+    chain_idx_key="chain_indices",
+    cdr3_col: str = "junction_aa",
+    *,
+    by: Sequence[Literal["gene_segment", "clonotype", "length"]] = "length",
+    target_col: Union[None, str] = None,
+    gene_annotation: Union[None, list] = None,
+    clonotype_id: Union[None, list] = None,
+    clonotype_key: Union[None, str] = None,
+    cdr_len: int,
+    plot: bool = True,
```
grst, line 30: I think we don't need that option, better to always use matplotlib. If the user wants to save a logo, they can do that through matplotlib.

grst (commented):
Regarding the function signature, I'm not a big fan of one function that tries to do different things by accepting different, mutually exclusive parameters. 

I'd prefer to split it up and have 
```
logoplot_cdr3_motif_length
logoplot_cdr3_motif_gene_segment
logoplot_cdr3_motif_clonotype
```

where each has only the corresponding parameters. If you want to reuse code between those functions, you can do so by factoring it out into a helper function. 

Alternatively, the function could just make a motif of all sequences in the anndata object and we could leave filtering to the user entirely, e.g. 

```
# clonotype
logoplot_cdr3_motif(mdata[mdata.obs["clone_id"] == "42", :]
# length
logoplot_cdr3_motif(mdata[ir.get.airr(mdata, "VJ_1", "junction_aa").str.len() == 15, :])
...
```

We could also have a combintation of the two where there's one implementation that makes a plot with all sequences in the AnnData object, and the `length`, `clonotype` and `gene_segment` versions are wrappers around the former that do the filtering for the user. 

What do you think?

MKanetscheider:
> Regarding the function signature, I'm not a big fan of one function that tries to do different things by accepting different, mutually exclusive parameters.
> 
> I'd prefer to split it up and have
> 
> ```
> logoplot_cdr3_motif_length
> logoplot_cdr3_motif_gene_segment
> logoplot_cdr3_motif_clonotype
> ```
> 
> where each has only the corresponding parameters. If you want to reuse code between those functions, you can do so by factoring it out into a helper function.
> 
> Alternatively, the function could just make a motif of all sequences in the anndata object and we could leave filtering to the user entirely, e.g.
> 
> ```
> # clonotype
> logoplot_cdr3_motif(mdata[mdata.obs["clone_id"] == "42", :]
> # length
> logoplot_cdr3_motif(mdata[ir.get.airr(mdata, "VJ_1", "junction_aa").str.len() == 15, :])
> ...
> ```
> 
> We could also have a combintation of the two where there's one implementation that makes a plot with all sequences in the AnnData object, and the `length`, `clonotype` and `gene_segment` versions are wrappers around the former that do the filtering for the user.
> 
> What do you think?

Actually, I was also not a big fan of my solution here, but I wasn't sure if including several similar functions is preferable. I will adapt it accordingly once I rewrote the code so that it properly returns matplotlib plots instead of SVG.

I will experiment a little bit, but I am afraid that it's not possible to plot a logo of all sequences as they need to be aligned and this is just the case for junction sequences that have the same length.

pyproject.toml
MKanetscheider: I am just having a hard time to manipulate the plot as plt.show() is called inside mpl_plot and it returns three objects, which are somewhat confusing to me...

Do you know if a object such as 
[[(Text(0, 0, 'C'), 178.8135375281251)],
 [(Text(0, 0, 'A'), 178.8135375281251)],
 [(Text(0, 0, 'S'), 178.8135375281251)],
 [(Text(0, 0, 'S'), 178.8135375281251)],
 [(Text(0, 0, 'E'), 11.495650387247952),
  (Text(0, 0, 'H'), 11.495650387247952),
  (Text(0, 0, 'L'), 11.495650387247952),
  (Text(0, 0, 'T'), 11.495650387247952),
  (Text(0, 0, 'Y'), 11.495650387247958),
  (Text(0, 0, 'P'), 22.991300774495915)],
 [(Text(0, 0, 'G'), 13.08748712194183),
  (Text(0, 0, 'S'), 13.08748712194183),
  (Text(0, 0, 'Y'), 13.08748712194183),
  (Text(0, 0, 'P'), 26.17497424388366),
  (Text(0, 0, 'V'), 26.17497424388366)],
 [(Text(0, 0, 'G'), 11.495650387247952),
  (Text(0, 0, 'I'), 11.495650387247952),
  (Text(0, 0, 'L'), 11.495650387247952),
  (Text(0, 0, 'S'), 11.495650387247952),
  (Text(0, 0, 'V'), 11.495650387247958),
  (Text(0, 0, 'W'), 22.991300774495915)],
 [(Text(0, 0, 'F'), 17.862997326023464),
  (Text(0, 0, 'A'), 35.72599465204693),
  (Text(0, 0, 'G'), 71.45198930409386)],
 [(Text(0, 0, 'A'), 13.688315950194387),
  (Text(0, 0, 'E'), 13.688315950194387),
  (Text(0, 0, 'L'), 13.688315950194387),
  (Text(0, 0, 'V'), 13.688315950194387),
  (Text(0, 0, 'G'), 41.06494785058316)],
 [(Text(0, 0, 'A'), 11.49565038724795),
  (Text(0, 0, 'E'), 11.49565038724795),
  (Text(0, 0, 'L'), 11.49565038724795),
  (Text(0, 0, 'P'), 11.49565038724795),
  (Text(0, 0, 'V'), 11.495650387247956),
  (Text(0, 0, 'G'), 22.9913007744959)],
 [(Text(0, 0, 'T'), 14.679323856635706),
  (Text(0, 0, 'G'), 29.358647713271413),
  (Text(0, 0, 'L'), 29.358647713271413),
  (Text(0, 0, 'P'), 29.358647713271402)],
 [(Text(0, 0, 'D'), 13.08748712194183),
  (Text(0, 0, 'I'), 13.08748712194183),
  (Text(0, 0, 'L'), 13.08748712194183),
  (Text(0, 0, 'G'), 26.17497424388366),
  (Text(0, 0, 'S'), 26.17497424388366)],
 [(Text(0, 0, 'N'), 15.280152684888265),
  (Text(0, 0, 'I'), 15.280152684888265),
  (Text(0, 0, 'L'), 30.56030536977653),
  (Text(0, 0, 'S'), 45.840458054664815)],
 [(Text(0, 0, 'L'), 17.472818247834695),
  (Text(0, 0, 'G'), 52.41845474350409),
  (Text(0, 0, 'S'), 52.41845474350409)],
 [(Text(0, 0, 'N'), 13.688315950194387),
  (Text(0, 0, 'Q'), 13.688315950194387),
  (Text(0, 0, 'S'), 13.688315950194387),
  (Text(0, 0, 'T'), 13.688315950194387),
  (Text(0, 0, 'A'), 41.06494785058316)],
 [(Text(0, 0, 'D'), 13.688315950194387),
  (Text(0, 0, 'Q'), 13.688315950194387),
  (Text(0, 0, 'E'), 13.688315950194387),
  (Text(0, 0, 'Y'), 13.688315950194387),
  (Text(0, 0, 'N'), 41.06494785058316)],
 [(Text(0, 0, 'E'), 15.280152684888265),
  (Text(0, 0, 'P'), 15.280152684888265),
  (Text(0, 0, 'T'), 30.56030536977653),
  (Text(0, 0, 'V'), 45.840458054664815)],
 [(Text(0, 0, 'L'), 60.16698866690969), (Text(0, 0, 'Q'), 80.2226515558796)],
 [(Text(0, 0, 'H'), 17.472818247834695),
  (Text(0, 0, 'T'), 52.41845474350409),
  (Text(0, 0, 'Y'), 52.41845474350409)],
 [(Text(0, 0, 'F'), 178.8135375281251)]]
can be easily transformed into a matplotlib figure or if it can be used to customize the returned matplotlib plot?

grst: > as plt.show() is called inside mpl_plot 

that's annoying. 

When choosing palmotif, did you also take a look at logomaker?
https://logomaker.readthedocs.io/en/latest/

Neither palmotif nor logomaker seem very actively maintained, but logomaker seems much more popular (according to github stars). From a first glance at the docs, logomaker seems more customizable and I also like that it doesn't have parasail as a hard dependency which I'd like to get rid of soon (see #450). But I'm not sure if it has other limitations.

MKanetscheider: To be honest, I am not sure why I have chosen palmotif...probably because it was used in the single-cell best practice book as well. I will have a look on logomaker, but if it's even preferable to use logomaker over palmotif (to get rid of parasail) I would be happy to adapt my code accordingly unless it is for whatever reason impossible to use logomaker in our case...

grst:
> but I am afraid that it's not possible to plot a logo of all sequences as they need to be aligned and this is just the case for junction sequences that have the same length.

Unless you use the hamming distance, this wouldn't even be guaranteeded for a clonotype I think? I see two options
 * check if all sequences are of the same length and fail otherwise
 * perform a multiple-sequence alignment (there must be python packages, but I haven't used any so far) before generating the logo plot.

grst:
> perform a multiple-sequence alignment (there must be python packages, but I haven't used any so far) before generating the logo plot.

this might be something that palmotif is doing for us (while logomaker does not) and the reason why it requires parasail.

MKanetscheider:
Hi there!
So after spending some time getting to know logomaker I decided that it's actually way easier to code with and offers better customization opportunities. Further, after you stated that you want to get rid of parasail as fast as possible, I think there is really no reason any more to still use palmotif. 

So what I have done on my latest commit was that I rewrote the function according to the logomaker synthax/workflow and simplified it a lot. There is now no conditional syntax any more that applies a filter function on the Anndata object for the user. I thought a lot about it and made the decision that it's probably better to leave filtering entirely to the user (as suggested by you in a previous comment) as it's not that difficult and any user can do as he/she prefers.

Multiple sequence alignment was not possible in palmotif either and I am not sure if it's even useful for cdr3 sequences. In all the paper that I have encountered so far, they generate CDR3 motifs only from sequences of the same length and thus MSA was never performed. This function too expects sequences to be the same length and raises and error if that's not the case. Regarding filter approaches, I'd like to showcase how easy it's done in the upcoming tutorial :)

The only thing that bothers me is a future warning raised while using logomaker.alignment_to_matrix() as I am not sure if this will become a serious issue in the (near) future:
image

Please let me know what you think about this warning and my "new" approach to this function :)

grst:
> The only thing that bothers me is a future warning raised while using logomaker.alignment_to_matrix() as I am not sure if this will become a serious issue in the (near) future:

This will become an issue with future pandas releases. There's even an issue here: 
https://github.com/jbkinney/logomaker/issues/36

Unfortunately, logomaker seems quite unmaintained (last commit 5 years ago), but I couldn't find any better alternatives. 
I suggest we still move forward with it. You can suppress the warning as described here: https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings

If the package really breaks eventually, we can still consider copying the code to scirpy, or forking the repo to scverse or something like that.

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,89 @@
...
+    cdr3_col: str = "junction_aa",
+    to_type: Sequence[Literal["information", "counts", "probability", "weight"]] = "information",
+    pseudocount: float = 0,
+    background=None,
+    center_weights: bool = False,
+    plot_default=True,
+    **kwargs,
+):
+    """
+    A user friendly wrapper function for the logomaker python package.
```
grst, line 26:
```suggestion
    Make Logoplots of CDR3 sequences.

    This is a wrapper around the logomaker package.
```

First line should always be a concise summary of what a function does, followed by a blank line.

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,89 @@
...
+    Enables the analysis of potential amino acid motifs by displaying logo plots.
+    Subsetting of AnnData/MuData has to be performed manually beforehand (or while calling) and only cells with equal cdr3 sequence lengths are permitted.
+
+    Parameters
+    ----------
+    {adata}
+    chains
+        One or multiple chains from which to use CDR3 sequences
```
grst, line 34: Not sure it makes sense to mix VJ/VDJ chains in a single plot. It could however make sense to mix VJ1/VJ2 etc

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,89 @@
...
+    chains: Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"] | Sequence[Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"]] = "VDJ_1",
+    airr_mod="airr",
+    airr_key="airr",
+    chain_idx_key="chain_indices",
+    cdr3_col: str = "junction_aa",
+    to_type: Sequence[Literal["information", "counts", "probability", "weight"]] = "information",
+    pseudocount: float = 0,
+    background=None,
+    center_weights: bool = False,
+    plot_default=True,
```
grst, line 22: I would just set defaults and allow the user to override them via kwargs. This could be done via dict.update()

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,89 @@
...
+    Subsetting of AnnData/MuData has to be performed manually beforehand (or while calling) and only cells with equal cdr3 sequence lengths are permitted.
+
+    Parameters
+    ----------
+    {adata}
+    chains
+        One or multiple chains from which to use CDR3 sequences
+    {airr_mod}
+    {airr_key}
+    {chain_idx_key}
```
grst, line 37: We typicalle allow the user to specify an `ax` object into which the plot is added. This allows the user to easily compose multi-panel plots, e.g.

```python
fig, ax =plt.subplots()
ir.pl.something (..., ax=ax)
```

Do you think this is possible with logomaker?

grst:
I've seen one example in a paper where they 
* put logoplots of VJ / VDJ seqs next to each other
* also made logos of VDJ gene usage.

See https://github.com/scverse/scirpy/issues/12#issuecomment-616574506

What do you think of these ideas?

MKanetscheider:
> I've seen one example in a paper where they
> 
> * put logoplots of VJ / VDJ seqs next to each other
> * also made logos of VDJ gene usage.
> 
> See [#12 (comment)](https://github.com/scverse/scirpy/issues/12#issuecomment-616574506)
> 
> What do you think of these ideas?

I have also encountered this many times and I think it would be a nice functionality as the information usually complements each other quite well. However, I am not sure if and how it could work with our current logomaker implementation

src/scirpy/pl/_logoplots.py
MKanetscheider: So if I understand it right, you want me to exclude them as part of the function call (like they are now) and define them in the function body with possibility to overwrite them as part of kwargs?

src/scirpy/pl/_logoplots.py
MKanetscheider: I see.
I will have a look on your other graphical implementations, but as far as I'm concerned (by looking at the logomaker documentation with working examples) this should be possible. So the idea is that the user can access the ax object after using the function right? So it has to be returned at some point in the function call, right?

grst: it's mostly useful for figure composition, i.e. one can make a multi planel figure using
```python
fig, (ax1, ax2) = plt.subplots(2,1)
```

and then plot different things into `ax1` and `ax2`.

src/scirpy/pl/_logoplots.py
grst: I thought about it again: maybe even better to just put them as additional function arguments... like that they are clearly documented and can easily be overridden. 
Maybe we can get away without specifying a font, see also my comment below.

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,91 @@
...
+from logomaker import Logo, alignment_to_matrix
+
+from scirpy.get import airr as get_airr
+from scirpy.util import DataHandler
+
+
+@DataHandler.inject_param_docs()
+def logoplot_cdr3_motif(
+    adata: DataHandler.TYPE,
+    chains: Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"] | Sequence[Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"]] = "VDJ_1",
```
grst, line 13:
```suggestion
    *, 
    chains: Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"] | Sequence[Literal["VJ_1", "VDJ_1", "VJ_2", "VDJ_2"]] = "VDJ_1",
```

Use keyword-only arguments here, to ensure that the order of all arguments except the two most central ones does not matter.

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,91 @@
...
+    airr_mod="airr",
+    airr_key="airr",
+    chain_idx_key="chain_indices",
```
grst, line 16:
```suggestion
    airr_mod: str="airr",
    airr_key: str="airr",
    chain_idx_key: str="chain_indices",
```

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,91 @@
...
+    cdr3_col: str = "junction_aa",
+    to_type: Sequence[Literal["information", "counts", "probability", "weight"]] = "information",
+    pseudocount: float = 0,
+    background=None,
```
grst, line 20: can we have a type hint for that? probably it's a np.ndarray?

MKanetscheider: I think you're referring to background here, right? The following screenshot is from logomaker api:
image
It seems that it accepts both np.ndarray or pd.dataframe...how can I typeset this accordingly? 💭

src/scirpy/pl/_logoplots.py
```diff
@@ -0,0 +1,91 @@
...
+        for sequence in airr_df[chain + "_" + cdr3_col]:
+            if sequence is not None:
+                sequence_list.append(sequence)
+
+    motif = alignment_to_matrix(
+        sequence_list, to_type=to_type, pseudocount=pseudocount, background=background, center_weights=center_weights
+    )
+    if plot_default:
+        cdr3_logo = Logo(
+            motif, font_name="Arial Rounded MT Bold", color_scheme="chemistry", vpad=0.05, width=0.9, **kwargs
```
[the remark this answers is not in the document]
MKanetscheider: No, there was no particular reason why I choose this font besides for aesthetic...didn't think that something like that would matter 🫨 
For now, I just removed the argument as I couldn't find out how exactly logomaker is inferring its default font-setting....maybe we should just point out in the function description that there are many different fonts? If you ran logomaker.list_font_names() one would see all possible fonts from which one could choose...is there an elegant way to combine it maybe with the function call like say a prompt so that the user has to decide a font while calling or something similar?

src/scirpy/pl/_logoplots.py
grst:
```suggestion
    background: np.ndarray | pd.DataFrame =None,
```

src/scirpy/pl/_logoplots.py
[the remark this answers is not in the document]
grst: > `logomaker.list_font_names()`

That's great, wasn't aware of this! I'd just stick with the default font and mention it in the documentation.

grst:
Thanks @MKanetscheider, I think I can take it from here! I'll apply some final cosmetic touches and then merge it.

PR #536, opened 2024-08-09
Mutational load function (SHM)

MKanetscheider:
Added mutational_load function to calculate differences between sequence and germline alignment. This is especially useful/insightful for BCR due to SHM and help to understand how much mutational actually occurred. However, this is a rather simple approach!

Closes #...

-   [ ] CHANGELOG.md updated


src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,278 @@
...
+    else:
+        for i, c in enumerate(germline):
+            if (c == ".") or (c == "N"):
+                continue
+            if (sequence[i] == ".") or (sequence[i] == "N"):
+                continue
+            elif c != sequence[i]:
+                distance += 1
+
+        return distance
```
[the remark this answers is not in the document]
felixpetschko: I think my implementation for the hamming distance might not be ideal in this case, since it's meant to create a matrix to compare all sequences to each other. The simple_hamming_distance however is mainly used to compare (a list of) pairs and also needs to ignore certain characters. This use case scales linearly with the input size, so maybe the performance is not a problem anyway.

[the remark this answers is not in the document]
grst: ok, then let's not optimize that for now. If it becomes a bottleneck we can think of applying some numba to the function or optimize it in any other way.

src/scirpy/tl/__init__.py
```diff
@@ -7,5 +7,6 @@
 from ._diversity import alpha_diversity
 from ._group_abundance import group_abundance
 from ._ir_query import ir_query, ir_query_annotate, ir_query_annotate_df
+from ._mutational_load import mutational_load
```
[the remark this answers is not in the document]
MKanetscheider: I added it and I think it looks quite good :)

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,278 @@
...
+                params.set_obs(f"{chain}_fwr4_mu_count", mutation_fwr4)
+
+            if not inplace and not frequency:
+                mutation_df[f"{chain}_fwr1_mu_count"] = mutation_fwr1
+                mutation_df[f"{chain}_cdr1_mu_count"] = mutation_cdr1
+                mutation_df[f"{chain}_fwr2_mu_count"] = mutation_fwr2
+                mutation_df[f"{chain}_cdr2_mu_count"] = mutation_cdr2
+                mutation_df[f"{chain}_fwr3_mu_count"] = mutation_fwr3
+                mutation_df[f"{chain}_cdr3_mu_count"] = mutation_cdr3
+                mutation_df[f"{chain}_fwr4_mu_count"] = mutation_fwr4
```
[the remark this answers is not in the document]
MKanetscheider: I rewrote the function based on this feedback and it looks now tidier! Please let me know if I should change/adapt it further

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,227 @@
...
+
+    if region == "subregion":
+        subregion_df = get_airr(params, ["fwr1", "fwr2", "fwr3", "fwr4", "cdr1", "cdr2", "cdr3"], chains)
+        for chain in chains:
+            airr_df[f"{chain}_junction_len"] = [len(a) for a in airr_df[f"{chain}_junction"]]
+
+            mutation_dict = {"fwr1": [], "fwr2": [], "fwr3": [], "fwr4": [], "cdr1": [], "cdr2": [], "cdr3": []}
+
+            for row in range(len(airr_df)):
+                fwr1_germline = airr_df.iloc[row].loc[f"{chain}_{germline_alignment}"][:78]
```
grst, line 151: Where do the numbers of the indices come from? Can we be sure they will remain stable?

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,227 @@
...
+                    )
+                )
+                mutation_dict["fwr4"].append(
+                    simple_hamming_distance(
+                        subregion_df.iloc[row].loc[f"{chain}_fwr4"],
+                        fwr4_germline,
+                        frequency=frequency,
+                        ignore_chars=ignore_chars,
+                    )
+                )
```
grst, line 218: I think this could be further simplified by 

(1)
defining a dict of regions
```python
regions = {
  "fwr1": (0, 78),
  "cdr1": (78, 114),
  ...
}
```

and then looping through it; somewhat like

```python
mutation_dict = {}
for region, coordinates in regions.items():
    mutation_dict[region] = simple_hamming_distance(
       subregion_df.iloc[row].loc[f"{chain}_fwr3"],
       airr_df.iloc[row].loc[f"{chain}_{germline_alignment}"][slice(*coordinates)]
       ...
   )
```

src/scirpy/tl/_mutational_load.py
MKanetscheider: These indices come from the IMGT unique numbering scheme (https://pubmed.ncbi.nlm.nih.gov/12477501/). This scheme is a standard approach to ensure that we can compare different V-regions of different cells. The neat thing is that sequences are aligned in a way that fwr 1-3 and cdr1-2 are always on the same spot in the germline and sequence alignment that's why these fixed indices work. cdr3 and fwr4 can be inferred by knowing the junction length and total sequence length as it is used in my code.

src/scirpy/tl/_mutational_load.py
MKanetscheider: I see :D
Thank's for the suggestions! I will try to further simplify it!

MKanetscheider: Thanks for the hint. I just adapted the code based on this suggestion and now it looks so much better :D

What's still pending is a robust test-case. I had a look on how other tests are written and I get the idea. However, I still don't know what's the best way to get/generate test-data. Should I manually generate a small (e.g. 10 sequences) dataset inside the test-function, which is used to test mutational_load or what is the best-practice here?
The dataset needs to be IMGT numbered so I don't think that I could load any of scirpy's native datasets here...

grst: >  Should I manually generate a small (e.g. 10 sequences) dataset inside the test-function

Yes, this is common practice. You can also put a small data file in `src/scirpy/tests/data`.

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,181 @@
...
+                    "cdr2": (165, 195),
+                    "fwr3": (195, 312),
+                    "cdr3": (312, 312 + airr_df.iloc[row].loc[f"{chain}_junction_len"] - 6),
+                    "fwr4": (
+                        312 + airr_df.iloc[row].loc[f"{chain}_junction_len"] - 6,
+                        len(airr_df.iloc[row].loc[f"{chain}_{germline_alignment}"]),
+                    ),
+                }
+
+                for v, coordinates in regions.items():
```
grst, line 164:
```suggestion
                for region, coordinates in regions.items():
```

One letter loop variables should only be used if they follow certain conventions, e.g. `i/j/k` for counters in for loops, 
or `k, v` for key, value pairs from `dict.items()`. 

Since you use `v` for the dict key, this can be confusing and I suggest to use a "proper" variable name like region here.

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,181 @@
...
+                mutation_df[f"{chain}_v_segment_{frequency_string}"] = mutations
+                return mutation_df
+
+    if region == "subregion":
+        subregion_df = get_airr(params, ["fwr1", "fwr2", "fwr3", "fwr4", "cdr1", "cdr2", "cdr3"], chains)
+
+        for chain in chains:
+            airr_df[f"{chain}_junction_len"] = [len(a) for a in airr_df[f"{chain}_junction"]]
+
+            mutation_dict = {"fwr1": [], "fwr2": [], "fwr3": [], "fwr4": [], "cdr1": [], "cdr2": [], "cdr3": []}
```
grst, line 148:
```suggestion
            mutation_dict = defaultdict(list)
```

grst:
In terms of implementation, I think we're getting there :)
Still need to try it out myself to check if I like the overall workflow/interface when using this in a juptyer notebook.

src/scirpy/tl/_mutational_load.py
MKanetscheider: Oke, I'll come up with something for my next push here :)

MKanetscheider:
Hi Gregor,
I worked on the test case for the mutational_load function and finished some kind of "beta" version. I would be very grateful if you could have a look if I'm going in the right direction here :)
Additionally, I discovered some bugs while testing, which I quickly resolved, but maybe not elegantly so please have also a look on that.

For some reason pushing these changes seem to have broken something with MuData, but I have no idea why and what I could possibly have done to cause this 😢 The error massage seems to be everywhere the same:
ImportError: cannot import name 'AlignedViewMixin' from 'anndata._core.aligned_mapping'
Could you help me solve this?

grst:
Breaking mudata is not your fault. It was caused by an anndata release and should be fixed by now. Just rerun the tests :)

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,203 @@
...
+    ----------
+    {adata}
+    {airr_mod}
+    {airr_key}
+    chain_idx_key
+        Key to select chain indices
+    sequence_alignment
+        Awkward array key to access sequence alignment information
+    germline_alignment
+        Awkward array key to access germline alignment information -> best practice mask D-gene segment (https://genomemedicine.biomedcentral.com/articles/10.1186/s13073-015-0243-2)
```
grst, line 66: How do these information get there? Does dandelion / immcantation preprocessing populate these fields with "best practice" values? Does cellranger also provide them without any further processing?

grst:
@MKanetscheider, two more considerations

1. Can we somehow verify that the alignments are really using the IMGT reference? Could we at least check something like length? If they were using a different reference, then the results would be pretty nonsensical, wouldn't they? 
2. Especially in the case of region = "subregion", we end up with a ton on columns in `adata.obs`. To me it seems that the mutational load is actually a chain-level attribute, so how about adding it to the awkward array instead? In that case we could get rid of the following arguments
  
   * chain_idx_key (just calculate it for all chains)
   * chains (just calculate it for all chains)
   * region (just calculate it for all regions)
   * frequency (just calculate both)
  
    The function would then add the chain-level attributes `mutation_count`, `mutation_freq`, `{cdr,fwd}{1,2,3,4}_mutation_{count,load}` `v_segment_mutation_{count,freq}`. 
  
   Afterwards, they could be retrieved like any other chain attribute: 
  
    ```python
    df = ir.get.airr_df(mdata, ["VDJ_1"], ["mutation_count", "mutation_freq"])
    ```


LMK what you think!

src/scirpy/tl/_mutational_load.py
MKanetscheider: Yes, these columns are populated by such tools like Dandelion and Immcantation. In fact, it's the result of re-annotation with igBLAST or imgt/highv-quest that is run "under the hood" of these tools (see also https://immcantation.readthedocs.io/en/stable/getting_started/10x_tutorial.html#assign-v-d-and-j-genes-using-igblast). Both dandelion and Immcantation follow the AIRR Community Standard, meaning that both "sequence_alignment" and "germline_alignment" should always result in the IMGT-gapped sequence (see also https://immcantation.readthedocs.io/en/stable/datastandards.html)
As far as I am aware does cellranger not provide it in this format, which is the main reason why we have to re-annotate cellranger output in the first place 😢

MKanetscheider:
> @MKanetscheider, two more considerations
> 
> 1. Can we somehow verify that the alignments are really using the IMGT reference? Could we at least check something like length? If they were using a different reference, then the results would be pretty nonsensical, wouldn't they?

Unfortunately yes, if sequences are not IMGT aligned this whole function would be either non-functional any more or just returning some random nonsense 😢 I think I actually do already check for length, because I count differences via hamming-distance, which raises a ValueError if germline and sequence alignments have different lengths. However, I would also like to have a better safety net, but I couldn't come up with anything else so far...

> 2. Especially in the case of region = "subregion", we end up with a ton on columns in `adata.obs`. To me it seems that the mutational load is actually a chain-level attribute, so how about adding it to the awkward array instead? In that case we could get rid of the following arguments
>    
>    * chain_idx_key (just calculate it for all chains)
>    * chains (just calculate it for all chains)
>    * region (just calculate it for all regions)
>    * frequency (just calculate both)
>    
>    The function would then add the chain-level attributes `mutation_count`, `mutation_freq`, `{cdr,fwd}{1,2,3,4}_mutation_{count,load}` `v_segment_mutation_{count,freq}`.
>    Afterwards, they could be retrieved like any other chain attribute:
>    ```python
>    df = ir.get.airr_df(mdata, ["VDJ_1"], ["mutation_count", "mutation_freq"])
>    ```
> 
> LMK what you think!

I think this sounds rather amazing...as you are already well aware, this whole function is rather ugly and not that user-friendly at the moment...I would really love to see it in a more compact format
But do you think that we might run into any performance problems if we always calculate everything with one function call? The function would have to calculate every hamming distance of each sequence/germline alignment pair, which could be troublesome for this big datasets that Felix and you are trying to reach, right?

grst:
> I think I actually do already check for length, because I count differences via hamming-distance, which raises a ValueError if germline and sequence alignments have different lengths

I think that's good enough then. After all it's also clearly stated in the documentation. 

> But do you think that we might run into any performance problems if we always calculate everything with one function call? 

I don't think it would be an issue, because it will still be linear over the number of chains. The part Felix has been working on compares all-vs-all sequences, which is quadratic over the number of sequences and therefore a much harder problem.

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,203 @@
...
+
+    for l1, l2 in zip(sequence, germline, strict=False):
+        if l1 in ignore_chars or l2 in ignore_chars:
+            num_chars -= 1
+
+        elif l1 != l2:
+            distance += 1
+
+    if num_chars == 0:
+        return np.nan  # can be used as a flag for filtering
```
grst, line 27: So this basically returns None when no (non-ignored) characters were compared. Could you please elaborate what's the reasoning behind this instead of returning 0?

grst:
> I think this sounds rather amazing...as you are already well aware, this whole function is rather ugly and not that user-friendly at the moment...I would really love to see it in a more compact format
But do you think that we might run into any performance problems if we always calculate everything with one function call? The function would have to calculate every hamming distance of each sequence/germline alignment pair, which could be troublesome for this big datasets that Felix and you are trying to reach, right?

I gave it a try in https://github.com/scverse/scirpy/pull/573. 
Speed-wise, it is about as fast as running all regions on VJ_1/VDJ_1 chains in your implementation, but including all chains. It could also be further optimized, but it's likely not a bottleneck... it still completes within a few minutes on 1M cells. 

I'd still need to deal with a bunch of edge cases, but I think the approach is viable.

src/scirpy/tl/_mutational_load.py
```diff
@@ -0,0 +1,203 @@
...
+                        "fwr1": (0, 78),
+                        "cdr1": (78, 114),
+                        "fwr2": (114, 165),
+                        "cdr2": (165, 195),
+                        "fwr3": (195, 312),
+                        "cdr3": (312, 312 + airr_df.iloc[row].loc[f"{chain}_junction_len"] - 6),
+                        "fwr4": (
+                            312 + airr_df.iloc[row].loc[f"{chain}_junction_len"] - 6,
+                            len(airr_df.iloc[row].loc[f"{chain}_{germline_alignment}"]),
+                        ),
```
grst, line 181: Do you have a reference for this calculation? 

From [Lefranc 2023](https://pubmed.ncbi.nlm.nih.gov/12477501/):

> rearranged CDR3-IMGT of 13 amino acids (and toa JUNCTION of 15 amino acids, 2nd-CYS 104 and J-TRP or J-PHE 118 being included in the JUNCTION definition). This numbering is convenient to use since 80% of the IMGT/LIGM-DBIG and TR rearranged sequences have a CDR3 IMGT length less than or equal to 13 amino acids  (IMGT statistics, October 2001). For rearranged CDR3-IMGT less than 13 amino acids, gaps are created from the top of the loop, in the following order 111, 112, 110, 113, 109, 114, etc. (Table 4A). For rearranged CDR3-IMGT more than 13 amino acids, additional positions are created, between positions 111 and 112 at the top of the CDR3-IMGT loop, in the following order 112.1, 111.1, 112.2, 111.2, 112.3, 111.3, etc. (Table 4B). 


From that text, I'd rather deduce something like

```
312 + max(junction_length, 13 * 3) # * 3 because 13 amino acids into nucleotides
```

But I haven't looked at the actual data myself.

src/scirpy/tl/_mutational_load.py
MKanetscheider: I thought that it could be useful as a flag for filtering. If num_chars = 0 that essentially means that there is no (useful) sequence alignment available, meaning that the specific read isn't useful for downstream analysis. So I thought by returning np.nan instead, it would be possible to distinguish and filter those reads. If you returned 0, wouldn't that mean it looks like a perfect alignment i.e. that sequence and germline are identical?

src/scirpy/tl/_mutational_load.py
MKanetscheider: This is actually also inspired by shazam's setRegionBoundaries helper function (https://shazam.readthedocs.io/en/stable/topics/setRegionBoundaries/). What is important to keep in mind is that the junction itself always contains the first codon of fwr4 and the last codon of fwr3 and thus the length of the cdr3 sequence (in nt) is always junction length - 6.
If gaps are created i.e. the sequence is less than 13 aa long does that even change anything? I'm not sure if these gaps would be included into the calculation of junction length, but if they are the code should still calculate the accurate boundaries. However, I was not aware of the fact that the cdr3 region may contain gaps as I apparently didn't read that particular part of that paper 😢

grst: Thanks for the reference of the Shazam function! I'll look into their code and compare.

grst:
Closing in favor of #573. 
Thanks for all your input @MKanetscheider! The optimized version in #573 wouldn't have been possible without it!

MKanetscheider:
> Closing in favor of #573. Thanks for all your input @MKanetscheider! The optimized version in #573 wouldn't have been possible without it!

Thanks for the nice words, but to stay realistic my input was at best the ground idea, while my implementation was far of anything useful. Thank you so much that you still could extract something useful for this amazing package out of it, which hopefully contributes to research in various labs 😄

PR #541, opened 2024-08-19
Hamming distance implementation with numba.cuda (GPU)

felixpetschko:
Hamming distance implementation with numba.cuda for GPU support.
This is built on top of the changes in Hamming distance implementation with Numba #512


felixpetschko:
The current numba.cuda GPU implementation gives me a speedup of around 2 (10 seconds vs. 5 seconds) compared to the improved numba cpu implemenation in https://github.com/scverse/scirpy/pull/512 on my local machine (6 i7 cpu cores vs. nvidia geforce gtx 1050) for the WU2020 dataset. Maybe I can try it on the some stronger GPU machines in the next days if i manage to get access to them.

I think it's nice to have a GPU implementation regardless of the actual speedup, because some people might only have access to strong GPUs machines. It's also good to have the foundation for further GPU developments. If someone has a good idea for a fast GPU kernel they can be exchanged easily. 

What do you think about this approach @grst?

grst:
@felixpetschko could you please resolve the merge conflicts here? Then the diff would be a lot cleaner

felixpetschko:
> @felixpetschko could you please resolve the merge conflicts here? Then the diff would be a lot cleaner

Done. The important changes are in the class [GPUHammingDistanceCalculator](https://github.com/felixpetschko/scirpy/blob/57e31a66575baa196847e4a14d2822fd62b7e504/src/scirpy/ir_dist/metrics.py#L757)

src/scirpy/ir_dist/__init__.py
```diff
@@ -60,6 +60,10 @@ def IrNeighbors(*args, **kwargs):
         BLOSUM62 matrix. Faster implementation of `alignment` with some loss.
         This option is incompatible with nucleotide sequences.
         See :class:`~scirpy.ir_dist.metrics.FastAlignmentDistanceCalculator`.
+      * `normalized_hamming` -- Normalized Hamming distance (in percent) for CDR3 sequences of equal length.
+        See :class:`~scirpy.ir_dist.metrics.HammingDistanceCalculator`.
+      * `tcrdist` -- Distance based on pairwise sequence alignments between TCR CDR3 sequences based on the tcrdist metric.
+        See :class:`~scirpy.ir_dist.metrics.TCRdistDistanceCalculator`.
```
grst, line 66:
```suggestion
```

this seems duplicated now

grst (commented):
Thanks, looks good, but I really can't judge the GPU code tbh... I was also hoping that with numba more code would be shared, but it really seems like it requires a completely different implementation to be optimized for GPU -- but I guess it would have been too good to be true. 

The real question for me is how to properly set this up and test this. 
 - Either enable GPU CI for scirpy (it's possible, scverse has this for scvi-tools and for rapids-singlecell, but it costs some money)
 - Move the GPU code to a separate repo (maybe even rapids-singlecell). 

Looping in @Intron7, would be great to hear your thoughts about this (and also about the numba GPU code in this PR).

grst:
@felixpetschko, let's discuss details with @Intron7 at the scverse conference

Intron7:
Hey @felixpetschko I looked at your kernel and they technically work, but you can make them even better. I would advocate for using prefetching and also to use one block for each `seqs_L1`. I have to look a bit more in how numba does cuda because I usually use cupy for kernels. You can also do a thread-wise comparison. This would lead to less warp divergence and more throughput. Lets discuss this further at scverse-conference on tuesday.

felixpetschko:
Hey @Intron7, thanks a lot for your feedback! There are some optimization possibilties for sure, because I am quite new to GPU programming. I am looking forward to seeing you at the scverse conference!

grst:
Hi @felixpetschko, what's the status here? Do you need anything from myself or Severin? 

I've seen you switched to Cupy, could you elaborate how that compares to the numba implementation?

felixpetschko:
Hi @grst! I am mainly done with my implementation here. Currently the speedup on my laptop for 1 million cells for the ir_dist function with hamming is at around 10 (45 vs. 480 seconds) compared to the new fast numba CPU implementation (and probably >100 compared to the original CPU implemenation). I think this is also the maximum speedup I would aim at for now, because there are currently some sequential parts (1 cpu) in the ir_dist function besides the hamming GPU kernel and the upstream processing for reading and preparing the data takes already longer anyway. So further optimization of the hamming kernel wouldn't be very effective.

My plan would be to prepare a pull request that is ready for review over the next days.

The reasons for switching to CuPy were the following:
Numba cuda only provides limited cuda features and you never really know how the numba code is mapped to the cuda features internally. However CuPy allows you to write the cuda kernels directly in C++ and offers all cuda features (that i have seen so far). Also most online ressources about GPU programming are about cuda kernels written in C++ and numba cuda is very niche. Also other programmers that might look at the code in the future will probably only know C++ cuda kernels (if they already did GPU programming). When doing GPU programming I use profiling tools to find out what is actually going on at the hardware and what the compiler did, so having this additional abstraction level with numba can actually be a nightmare.

grst:
Makes sense, thanks! 
Still not sure if this should be merged in to scirpy or into rapids-singlecell then. @Intron7, I'll also bring that up at the next core dev meeting what's our long-term strategy here.

Intron7:
Hey @felixpetschko can you send me a larger dataset to test this? I have some ideas and want to see if this works.

grst:
> Still not sure if this should be merged in to scirpy or into rapids-singlecell then. @Intron7, I'll also bring that up at the next core dev meeting what's our long-term strategy here.

@felixpetschko, the outcome of this discussion was that the function stays here, and we'll setup a GPU CI for scirpy. @ilan-gold  or @flying-sheep can help with that once this PR is ready.

grst:
@Intron7, Appears phil created documentation for the gpu ci: https://github.com/scverse/governance/blob/main/developer/gpu_ci.md

Maybe this helps?

Intron7:
@grst I talked to @Zethson and he adjusted the setting in cirun. The test is now running but failing.

src/scirpy/ir_dist/metrics.py
```diff
@@ -752,6 +752,345 @@ def _nb_hamming_mat():
...
+    gpu_n_blocks:
+        Number of blocks in which the final result matrix should be computed. Each block reserves GPU memory
+        in which the computed result block has to fit in sparse representation. Lower values give better performance
+        but increase the risk of running out of reserved memory. This value should be chosen based on the
+        estimated sparsity of the result matrix and the size of the GPU device memory.
+    gpu_block_width:
+        Maximum width of blocks in which the final result matrix should be computed. Each block reserves GPU memory
+        in which the computed result block has to fit in sparse representation. Higher values allow for a lower
+        number of result blocks (gpu_n_blocks) which increases the performances. This value should be chosen based on
+        the GPU device memory.
```
grst, line 775: Why do we need both? Can't you infer one based on the other?

src/scirpy/ir_dist/metrics.py
```diff
@@ -752,6 +752,345 @@ def _nb_hamming_mat():
...
+        n_blocks = self.gpu_n_blocks
+
+        seqs_mat2_blocks = np.array_split(seqs_mat2, n_blocks)
+        seqs_L2_blocks = np.array_split(seqs_L2, n_blocks)
+        seqs2_original_indices_blocks = np.array_split(seqs2_original_indices, n_blocks)
+        result_blocks = [None] * n_blocks
+
+        block_offset = start_column
+
+        print(
```
grst, line 1065: I suggest to use `logging.info` instead of `print`

src/scirpy/ir_dist/metrics.py
```diff
@@ -752,6 +752,345 @@ def _nb_hamming_mat():
...
+            for i, seq in enumerate(seqs):
+                mat[i][0 : len(seq)] = np.frombuffer(seq.encode("ascii"), dtype=np.uint8)
+                L[i] = len(seq)
+            return mat, L
+
+        try:
+            seqs_mat1, seqs_L1 = _seqs2mat_fast(seqs, max_len=max_seq_len)
+            seqs_mat2, seqs_L2 = _seqs2mat_fast(seqs2, max_len=max_seq_len)
+        except UnicodeError:
+            print(
```
grst, line 858: I suggest to use logging.info instead of print

src/scirpy/ir_dist/metrics.py
felixpetschko: No, one can't be inferred based on the other.

Let's say we have a 1000x1000 (dense represenation) not yet computed result matrix with gpu_n_blocks=10 and gpu_block_width=20. Then the result matrix is computed in 10 blocks of  1000x100 (dense representation). Each of those blocks needs to fit into a 1000x20 block in SPARSE representation once computed and this 1000x20 block needs to fit into GPU memory. So there shouldn't be a resulting row in a block that has more than 20 values <= cutoff.

The parameter gpu_block_width should be chosen based on the available GPU memory. Choosing lower values for gpu_n_blocks increases the performance but also increases the risk of running out of reserved memory, since the result blocks that need to fit into the reserved GPU memory in sparse representation get bigger.

(I will add this explanation to the documentation of the class)

In reality, the result matrix will most likely be very sparse. E.g. for 1 million cells I use gpu_n_blocks=10 and gpu_block_width=1100.

We could compute the maximum possible value for gpu_block_width if we know the GPU memory. But reading out hardware ressources programmatically can lead to troubles when working at a cluster or in virtual environemnts.

grst (approved):

src/scirpy/ir_dist/metrics.py
Intron7: Did you every think about creating a COO matrix and let this be compressed into CSR. Might this be easier and better?

Intron7:
For a large Matrix in the Hamming-Kernel have you checked that int is enough to cover the indexing of data?

src/scirpy/ir_dist/metrics.py
felixpetschko: I also thought about this option, but finally the current implementation with directly constructing the CSR matrix was the fastest solution I could come up with. Since the Hamming distance between two sequences can be computed very fast, being very fast at secondary tasks like the sparse matrix construction is very important in this case.

felixpetschko:
> For a large Matrix in the Hamming-Kernel have you checked that int is enough to cover the indexing of data?

That's a good point, but I think for our purposes int should be enough for the kernels. I think the most problematic part is the indptr array of the CSR matrices, therefore I added some int32 overflow checks. If such an overflow occures, the user can try it with a smaller cutoff to get less result values that need to be stored in the CSR matrix.

PR #557, opened 2024-09-21
Filter clonotype clusters to be displayed by anndata.obs data

felixpetschko:
Currently there is a problem with too crowded clonotype clustering result graphs. Therefore I was asked to implement something to filter the clonotype clusters that should be displayed in the result graph. My solution would be to allow the user to pass a list with cell indices of anndata.obs to the clonotype_network function. Only clusters that contain at least one of these cells should be shown. The call could look like this:

```python
condition = (adata.obs['gex:patient'] == 'Lung1') & (adata.obs['gex:source'] == 'Tumor')
cell_index_filter = adata.obs.index[condition].tolist()
ir.tl.clonotype_network(adata, min_cells=1, sequence="aa", metric="hamming", cell_index_filter=cell_index_filter)
_ = ir.pl.clonotype_network(adata, color="gex:patient", label_fontsize=9, panel_size=(7, 7), base_size=20)
```

image

That way the user would be able to do quite flexibel filtering which could be interesting for result graphs of any size. If this filtering would be done already at the beginning of the pipeline (before ir_dist) you couldn't see the whole clusters the filtered cells are part of. What do you think about this approach @grst ?

### DIFF
--- a/src/scirpy/tl/_clonotypes.py
+++ b/src/scirpy/tl/_clonotypes.py
@@ -543,7 +543,7 @@ def clonotype_network(
     if mask_obs is not None:
         cell_mask = adata.obs[mask_obs]
         cell_indices_reversed = {v: k for k, values in cell_indices.items() for v in values}
-        clonotype_mask = np.array([False] * len(cell_indices.items()))
+        clonotype_mask = np.zeros((len(cell_indices), ), dtype=bool)
         cell_index_filter = adata.obs.loc[cell_mask].index
         for cell_index in cell_index_filter:
             if cell_index in cell_indices_reversed:
--- a/src/scirpy/tl/_clonotypes.py
+++ b/src/scirpy/tl/_clonotypes.py
@@ -541,7 +541,7 @@ def clonotype_network(
 
     # create clonotype_mask for filtering according to mask_obs
     if mask_obs is not None:
-        if(isinstance(mask_obs, str)):
+        if isinstance(mask_obs, str):
             cell_mask = adata.obs[mask_obs]
         elif isinstance(mask_obs, np.ndarray) and mask_obs.dtype == np.bool_:
             cell_mask = mask_obs
--- a/src/scirpy/tl/_clonotypes.py
+++ b/src/scirpy/tl/_clonotypes.py
@@ -494,6 +494,10 @@ def clonotype_network(
     random_state
         Random seed set before computing the layout.
     {airr_mod}
+    cell_index_filter
+        List of cell indices in anndata obs to filter the clonotype clusters that should be displayed
+        in the graph. Only clusters that contain at least one cell of the cell_index_filter list will be shown.
+        Can be set to None to avoid filtering.
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


felixpetschko:
I also noticed something where I am not sure if it's a bug. In the function define_clonotype_clusters we allow 3 different clustering methods:
https://github.com/scverse/scirpy/blob/6ed2c9923ff7b88df6e80140aa0f74c306a9ed68/src/scirpy/tl/_clonotypes.py#L310-L319

But in clonotype_network we always use:
[`components = np.array(graph.decompose("weak")) `](https://github.com/scverse/scirpy/blob/6ed2c9923ff7b88df6e80140aa0f74c306a9ed68/src/scirpy/tl/_clonotypes.py#L534)

I think it might be necessary to also consider the clustering method in clonotype_network. Maybe this is already done but I couldn't find any code that does this.

grst:
> But in clonotype_network we always use:
components = np.array(graph.decompose("weak")) 

This was intentional. If multiple partitions were part of the same connected module, they could still be highlighted using different colors. My motivation for this was the following:
If we split up connected components
 - some clonotypes would be part of multiple modules and shown multiple times
 - we would artificially split up two modules that are actually connected in the graph. 

This doesn't mean that I'm against exploring layouts where this is solved differently.

grst:
Thanks, providing such a filtering strategy definitely makes sense. 

> My solution would be to allow the user to pass a list with cell indices of anndata.obs to the clonotype_network function.

I would prefer to pass a column in `adata.obs` or a boolean mask directly instead to a `mask_obs` argument. This is a convention adopted in scanpy already, e.g. 
https://github.com/scverse/scanpy/blob/a91bb02b31a637caeee77c71dcd9fbce8437cb7d/src/scanpy/preprocessing/_scale.py#L75

```python
adata.obs["filter_condition"] =  (adata.obs['gex:patient'] == 'Lung1') & (adata.obs['gex:source'] == 'Tumor')
ir.tl.clonotype_network(adata, min_cells=1, sequence="aa", metric="hamming", mask_obs="filter_condition")

# or
ir.tl.clonotype_network(
    adata, min_cells=1, sequence="aa", metric="hamming", 
    mask_obs=(adata.obs['gex:patient'] == 'Lung1') & (adata.obs['gex:source'] == 'Tumor')
)
```

felixpetschko:
Alright, I chose the first approach for my implementation.
```
adata.obs["filter_condition"] =  (adata.obs['gex:patient'] == 'Lung1') & (adata.obs['gex:source'] == 'Tumor')
ir.tl.clonotype_network(adata, min_cells=1, sequence="aa", metric="hamming", mask_obs="filter_condition")
```

If the current implementation is fine for you, I will add test cases and try it on the cluster.

src/scirpy/tl/_clonotypes.py
```diff
@@ -528,20 +533,40 @@ def clonotype_network(
...
     # store size in graph to be accessed by layout algorithms
-    clonotype_size = np.array([idx.size for idx in clonotype_res["cell_indices"].values()])
+    clonotype_size = np.array([len(idx) for idx in cell_indices.values()])
     graph.vs["size"] = clonotype_size
+
+    # create clonotype_mask for filtering according to mask_obs
+    if mask_obs is not None:
+        cell_mask = adata.obs[mask_obs]
+        cell_indices_reversed = {v: k for k, values in cell_indices.items() for v in values}
+        clonotype_mask = np.array([False] * len(cell_indices.items()))
```
grst, line 546: Not critical here, but this is more efficient ;)

```suggestion
        clonotype_mask = np.zeros((len(cell_indices), ), dtype=bool)
```

grst:
I actually think we should support both, a boolean mask and a column name. 

Implementation wise this looks fine!

felixpetschko:
> I actually think we should support both, a boolean mask and a column name.

Now I allow both options and added a test case. The worst case performance impact of the filtering should be neglectable, whereas reducing the number of connected modules in the graph with filtering can improve the performance.

grst (approved):
LGTM

PR #565, opened 2024-10-20
Add Stephenson datasets

grst:
Closes #...

-   [x] CHANGELOG.md updated

### DIFF
--- a/src/scirpy/datasets/__init__.py
+++ b/src/scirpy/datasets/__init__.py
@@ -28,7 +28,7 @@ HERE = Path(__file__).parent
 
 _FIGSHARE = pooch.create(
     path=pooch.os_cache("scirpy"),
-    base_url="doi:10.6084/m9.figshare.22249894.v1",
+    base_url="doi:10.6084/m9.figshare.22249894.v2",
     version=version("scirpy"),
     version_dev="main",
     env="SCIRPY_DATA_DIR",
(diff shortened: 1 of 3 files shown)


src/scirpy/datasets/__init__.py
```diff
@@ -124,6 +125,34 @@ def maynard2020() -> MuData:
...
+    and processed using the nf-core `RNA-seq pipeline <https://github.com/nf-core/rnaseq>`_ to obtain
+    gene expression and TraCeR/BraCeR to reconstruct receptors.
+
+    The processed files have been imported and transformed into an :class:`anndata.AnnData`
+    object using the following script:
+
+    .. code-block:: python
+
+        {processing_code}
+    """
```
grst, line 151: @MKanetscheider, could you please update the description here? 

For the other datasets, there are python scripts that specify exactly how the dataset was processed, e.g. https://github.com/scverse/scirpy/blob/main/src/scirpy/datasets/_processing_scripts/maynard2020.py that are included using the `processing_code` reference. 

If you don't have something like that, I'm also happy if you just roughly describe as text how the dataset was processed. Because, let's be honest, likely nobody will ever look at it. 

But at least citation and the reference to where you downloaded the data should be there :)

MKanetscheider: Hi I just wanted to push my changes to this PR, but apparently my access was denied.
image
Do you know if that's a problem on my side, or do you have to give some kind of permission? 🥲

grst: my bad, you should have received an invite now

MKanetscheider: No problem 😄 
For now, I just added a verbal description, because the code I used is kinda messy as I used to test a lot of different things in the same file... if at any point the code is required I will clean it up and provide here, of course!

PR #617, opened 2025-06-01
Improved result matrix stacking for Hamming GPU implementation

felixpetschko:
@grst Recently I tried the GPU implementation of the Hamming distance metric with 8 million cells of the Omniscope Covid dataset, which is the largest dataset that is currently available. I noticed performance problems with stacking the blocks of the final result matrix that are computed by GPU. Therefore I implemented a numba function to stack the blocks efficiently - stacking only takes 6 seconds now for 8 million cells with 1 CPU. 
With the new stacking implementation I was able to run 8 million cells in ~210 seconds with an Nvidia A30 GPU at the cluster with the GPU hamming metric (~2000 seconds with 64 CPUs with the CPU hamming metric).

### DIFF
--- a/.cruft.json
+++ b/.cruft.json
@@ -1,7 +1,7 @@
 {
   "template": "https://github.com/scverse/cookiecutter-scverse",
-  "commit": "83d178f1d000c547cc772368f82568bb72735d39",
-  "checkout": "test2",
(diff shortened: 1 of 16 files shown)


felixpetschko:
Hi @grst, I updated the changelog now

PR #680, opened 2026-03-08
Filter unused categories from clonotype network legend

### LINKED ISSUE #321: Remove unused labels from antigen-annotated clonotype networks
In antigen-annotated clonotype networks, we could have an option to remove from the legend the labels that are not present in the plot. It would make it easier to spot the correct color by eye without having too many, similar color levels in the palette.

LiudengZhang:
## Summary
- The clonotype network plot legend now only shows categories that are actually present in the plotted data
- Previously, all categories from the full dataset were listed, making it hard to identify colors when only a subset of categories appeared in the network
- Collects which category colors are used in the pie chart rendering, then filters the legend entries accordingly

Closes #321

## Test plan
- [ ] Verify legend only shows categories present in the plotted network subset
- [ ] Legend still works correctly when all categories are present
- [ ] No visual regression for non-categorical coloring (continuous variables, uniform color)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
+## [Unreleased]
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


grst (approved):
Thanks!

Issue #591, opened 2025-02-10
Option to use TCRblosum for distance calculation


riederd:
TCRblosum, a substitution matrix specifically derived from TCR sequences, is likely to provide a more accurate measure of TCR sequence similarity, however it is not yet used in scirpy.

I propose adding support for TCRblosum as an alternative substitution matrix for the distance calculation.  Ideally, this could be implemented as a parameter within the `ir_dist` function, allowing users to choose also TCRblosum.

**Publication:**

*   TCRblosum is described in [Postovskaya et al. (2025)](https://academic.oup.com/bib/article/26/1/bbae602/7906917)

grst:
It's not very well documented in scirpy, but it should be possible to define a substitution matrix in a file as supported by parasail:
https://github.com/jeffdaily/parasail-python?tab=readme-ov-file#substitution-matrices

and then use

```python
ir.pp.ir_dist(adata, metric="alignment", cutoff=10, subst_mat="tcrblosum.txt")
```

If that works, it would be quite easy to include it natively in scirpy.

riederd:
Missed that, will try and report back 
Thanks

grst:
Available in `dev` version via
```
ir_dist(metric="tcrdist", base_matric="tcrblosum")
```

Will be included in next release.

PR #685, opened 2026-04-06
Add tcrblosum support to TCRdist

felixpetschko:
So far, the TCRdist metric used a distance matrix derived from the blosum62 substitution matrix. This PR extends TCRdistDistanceCalculator with a new base_matrix="tcrblosum" option alongside the existing default blosum62 behavior. This way, distance matrices based on the tcrblosum substitution matrices (different matrices for alpha and beta chain) are used for the TCRdist metric calculation.

I try to illustrate how I derived the tcrblosum based distance matrices in this [google colab notebook](https://colab.research.google.com/drive/1K5UFs6bWWBJkN7l3BndkhJ2wGKKtHzZV?usp=sharing).

The usage of the tcrblosum matrices was already discussed in #591.


src/scirpy/ir_dist/metrics.py
```diff
@@ -1183,14 +1184,24 @@ class TCRdistDistanceCalculator(_MetricDistanceCalculator):
...
... ('T', 'A'): 4, ('T', 'R'): 4, ('T', 'N'): 4, ('T', 'D'): 4, ('T', 'C'): 4, ('T', 'Q'): 4, ('T', 'E'): 4, ('T', 'G'): 4, ('T', 'H'): 4, ('T', 'I'): 3, ('T', 'L'): 4, ('T', 'K'): 4, ('T', 'M'): 4, ('T', 'F'): 4, ('T', 'P'): 4, ('T', 'S'): 4, ('T', 'T'): 0, ('T', 'W'): 4, ('T', 'Y'): 4, ('T', 'V'): 4, ('W', 'A'): 4, ('W', 'R'): 4, ('W', 'N'): 4, ('W', 'D'): 4, ('W', 'C'): 4, ('W', 'Q'): 4, ('W', 'E'): 4, ('W', 'G'): 4, ('W', 'H'): 3, ('W', 'I'): 4, ('W', 'L'): 4, ('W', 'K'): 4, ('W', 'M'): 4, ('W', 'F'): 4, ('W', 'P'): 4, ('W', 'S'): 4, ('W', 'T'): 4, ('W', 'W'): 0, ('W', 'Y'): 4, ('W', 'V'): 4, ('Y', 'A'): 4, ('Y', 'R'): 4, ('Y', 'N'): 4, ('Y', 'D'): 4, ('Y', 'C'): 4, ('Y', 'Q'): 4, ('Y', 'E'): 4, ('Y', 'G'): 4, ('Y', 'H'): 4, ('Y', 'I'): 4, ('Y', 'L'): 4, ('Y', 'K'): 4, ('Y', 'M'): 4, ('Y', 'F'): 4, ('Y', 'P'): 4, ('Y', 'S'): 4, ('Y', 'T'): 4, ('Y', 'W'): 4, ('Y', 'Y'): 0, ('Y', 'V'): 4, ('V', 'A'): 4, ('V', 'R'): 4, ('V', 'N'): 4, ('V', 'D'): 4, ('V', 'C'): 4, ('V', 'Q'): 4, ('V', 'E'): 4, ('V', 'G'): 4, ('V', 'H'): 4, ('V', 'I'): 4, ('V', 'L'): 4, ('V', 'K'): 4, ('V', 'M'): 4, ('V', 'F'): 4, ('V', 'P'): 4, ('V', 'S'): 4, ('V', 'T'): 4, ('V', 'W'): 4, ('V', 'Y'): 4, ('V', 'V'): 0}
```
grst, line 1202: Did you make any modifications to the original TCRblosum matrix? e.g. the cap at 4?

grst (commented):
Implementation-wise this looks great! 

What's still missing is 
 * changelog update
 * Documentation-update of the user-facing (pp.ir_dist) method. Probably best to add a new metric `tcrblosum` or `tcrdist_tcrblosum`. 
 * Reference to the TCRblosum paper in the documentation
 * Maybe tutorial update?

src/scirpy/ir_dist/metrics.py
felixpetschko: First of all, you always have this conversion from the substitution matrix (blosum62, tcrblosum) to a distance matrix.
In the [tcrblosum paper](https://academic.oup.com/bib/article/26/1/bbae602/7906917?login=true) , they just menttion "Firstly, we transformed the tcrBLOSUM similarity matrix into a distance matrix according to the rules of TCRdist [[12]](https://www.nature.com/articles/nature22383)".

In the [original TCRdist paper](https://www.nature.com/articles/nature22383) they write:
"The mismatch distance is defined based on the BLOSUM62 (ref. 37) substitution matrix as follows: distance (a, a)=0; distance (a, b)=min (4, 4-BLOSUM62 (a, b)), where 4 is 1 unit greater than the most favourable BLOSUM62 score for a mismatch, and a and b are amino acids".

Now there would be two options:
1) also use the constant cap 4 like in the original tcrdist paper
2) use a cap that is one unit greater than the most favourable score

However, the most favourable score would be 1 for tcrblosum_alpha and 2 for tcrblosum_beta which would result in a quite low cap (see matrices in [notebook](https://colab.research.google.com/drive/1K5UFs6bWWBJkN7l3BndkhJ2wGKKtHzZV?usp=sharing)). Therefore I just went with the fixed cap of 4 for the transformation formula from substitution matrix to distance matrix of the original tcrdist implementation:
distance(a, a) = 0 and distance(a, b) = min(4, 4 - score)

Unfortunately, I wasn't able to find out how they did it exactly in the tcrblosum paper. Another thing to consider is, that the distance values for alpha and beta chain might be compared later during the clonotype clustering. Therefore having two different caps might cause problems. 

What do you think would be the best option?

grst: Dear @apostovskaya, 

we are working on integrating your tcrBLOSUM matrix into scirpy, which contains a reimplementation of the TCRdist algorithm. 

We are unsure how you intended the matrix to be used with TCRdist. Could you please clarify what you mean with

> Firstly, we transformed the tcrBLOSUM similarity matrix into a distance matrix according to the rules of TCRdist [[12]](https://www.nature.com/articles/nature22383)

in your paper and how you would suggest to set the cap?

grst: > "The mismatch distance is defined based on the BLOSUM62 (ref. 37) substitution matrix as follows: distance (a, a)=0; distance (a, b)=min (4, 4-BLOSUM62 (a, b)), where 4 is 1 unit greater than the most favourable BLOSUM62 score for a mismatch, and a and b are amino acids".

tbh, I never understood this part about TCRdist. It skews the matrix quite a bit, basically assigning a distance of 4 to all mismatches that have a negative distance in BLOSUM62 (even if it's just -1). So the only way to get a score $1 \leq s \leq 3$ is if you have one of the rare pairs with a positive mismatch score according to BLOSUM62. 

Having a cap of 2 with TCRblosum doesn't make sense to me, because then the strong negative effect of mismatches with `C` would be completely gone which defeats the purpose of TCRblosum. Using `4` is just as arbitrary.

felixpetschko:
I updated the approach now according to what was discussed via email. The distance matrices are now computed from the original substitution matrices (blosum62, tcrblosum_alpha, tcrblosum_beta) rather than stored as precomputed distance tables.

In general, the calculation is now done as follows:
`distance(a, a) = 0 and distance(a, b) = min(distance_cap, distance_offset - score)`

The `base_matrix(Literal["blosum62", "tcrblosum"] = "blosum62")` and the `distance_cap(int | None | Literal["default"] = "default")` parameters can be set in the calling `ir_dist` method. For blosum62 `distance_cap=4` by default and `distance_offset=4` as always. For the tcrblosum matrices, `distance_cap=None` (no cap) by default and `distance_offset=3`=(global maximum off-diagonal substitution score of both alpha and beta matrices +1). This distance offset maps the most favorable off-diagonal score across both tcrblosum matrices (2) to distance 1. That way, the blosum62 defaults follow the original TCRdist paper while the tcrblosum defaults should preserve the signal present in the more sparse tcrblosum matrices. 

My intention for the user call would be the following rather than creating an additional _tcrdist_tcrblosum_ metric:

_tcrdist with blosum62_
```python
ir.pp.ir_dist(
    mdata,
    metric="tcrdist",
    sequence="aa",
    cutoff=15,
    #base_matrix="blosum62", #set by default
    #distance_cap=4, #set by default
)
```

_tcrdist with tcrblosum matrices_
```python 
ir.pp.ir_dist(
    mdata,
    metric="tcrdist",
    sequence="aa",
    cutoff=15,
    base_matrix="tcrblosum",
    #distance_cap=None, #set by default
)
```

src/scirpy/ir_dist/metrics.py
```diff
@@ -343,23 +344,38 @@ def _compute_block(self, seqs1, seqs2, origin):
...
+        Maximum distance assigned to a mismatch. If `None`, mismatch distances are uncapped.
+    distance_offset:
+        Offset from which the substitution score is subtracted.
 
     Returns
     -------
     distance_matrix:
         distance lookup matrix
     """
     dm = np.zeros((len(alphabet), len(alphabet)), dtype=np.int32)
```
grst, line 370: Does this imply that letters in the alphabet, but not in the matrix have a distance of 0 (e.g. X vs. A)? 

Is this what we want? Or maybe rather Inf? 

FWIW, there's also variants of BLOSUM62 that include these extended amino acids: 

Image

src/scirpy/ir_dist/metrics.py
```diff
@@ -1213,6 +1323,49 @@ def __init__(
...
+            else:
+                raise ValueError("`chain_type` must be 'VJ' or 'VDJ' when `base_matrix='tcrblosum'`.")
+
+            off_diagonal = ~np.eye(len(self.matrix_alphabet), dtype=bool)
+            max_score = int(
+                max(
+                    np.max(self.tcrblosum_alpha_substitution_matrix[off_diagonal]),
+                    np.max(self.tcrblosum_beta_substitution_matrix[off_diagonal]),
+                )
+            )
```
grst, line 1358: Probably doesn't make a big difference, but do we have a rationale for computing a global max across both matrices, rather than doing this separately for each matrix?

grst:
Implementation-wise this looks good, and I also prefer to make the tcrdist implementation generic such that it works with arbitrary matrices rather than introducing another metric. 

We just need to make sure to advertise it appropriately in the tutorial, otherwise the functionality is a bit hidden.

src/scirpy/ir_dist/metrics.py
felixpetschko: That way the same substitution scores are mapped to the same distance values regardless of the chain type. This keeps VJ and VDJ distances on a shared scale, which is important later on during clonotype clustering, since VJ and VDJ distances should be comparable.

src/scirpy/ir_dist/metrics.py
felixpetschko: > Does this imply that letters in the alphabet, but not in the matrix have a distance of 0 (e.g. X vs. A)?

Yes - this behavior comes from the original pwseqdist base implementation. Maybe this was the pragmatic approach to treat unknown residues as "uninformative" rather than as evidence of dissimilarity. One could see capturing true matches as more important than avoiding false matches.

The question is if we want to change this.


> FWIW, there's also variants of BLOSUM62 that include these extended amino acids

However, probably we don't have something like that for the tcrblosum matrices.

src/scirpy/ir_dist/metrics.py
grst: ok, then let's keep it. Maybe add a line about it to the docstring for future reference.

felixpetschko:
> We just need to make sure to advertise it appropriately in the tutorial, otherwise the functionality is a bit hidden.

I made a short tutorial update to mention tcrblosum usage.

grst (approved):
LGTM now!

grst:
Sorry, apparently I forgot to merge this!

PR #688, opened 2026-04-13
drop distance_key from ir.tl.define_clusters

### LINKED ISSUE #687: distance_key is silently exchanged in define_clusters
### Report

I had trouble assiging the distance key in define_clonotypes, while digging for the issue I identified the issue (me!).

Although I still like to propose a change and can generate a pull request for it if one of my or another suggestion sounds good to you!

by defenition define_clonotypes, does not forward the distance_key to define_clonotype_clusters and thus does not allow for custom distance_keys here - as this is what define_clonotype_clusters does. 
Setting distance_key is still possible but is silently redefined in define_clonotypes, which I would like to change.

suggestions:
a) send a log info if distance_key != "ir_dist_nt_identity" that this will be used instead.
b) drop distance_key from the input parameters.
c) both of the above (if passed as a kwargs)

### Versions

```python

```

dansteiert:
- drop distance_key from input parameters
- drop distance_key from clause to potentially rerun ir.pp.ir_dist
- add unsilensing via logging warn for distance_key replacement.
- drop distance_key from kwargs if present
- pass distance key explicitly

Closes #687

- [ ] CHANGELOG.md updated
- [ ] Tests added (For bug fixes or new features)
- [ ] Tutorial updated (if necessary)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,13 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## v0.23.1
+## [Unreleased]
 
 ### Fixes
 
  - Filter unused categories from clonotype network legend ([#680](https://github.com/scverse/scirpy/pull/680/)).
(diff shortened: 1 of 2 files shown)


grst:
Thanks, I don't think the CI failures are yours. I'll look into it.

PR #723, opened 2026-07-07
Improve GPU Hamming distance computation

felixpetschko:
## Summary

This PR improves the GPU implementation of the Hamming distance metric by making the computation more memory-flexible and better suited for large datasets. The main change is that GPU Hamming computations can now be split into both row and column blocks. This replaces the previous single `gpu_n_blocks` parameter with `gpu_col_blocks` and the new `gpu_row_blocks` parameter.

Before, the result matrix was split only into column blocks using `gpu_n_blocks`, while `gpu_block_width` controlled how much GPU memory was reserved for result values below the cutoff. This already assumed a certain level of sparsity to avoid reserving unnecessary amounts of memory. For larger datasets, it was difficult to choose parameters that kept the number of blocks low enough for good performance while still fitting into GPU memory. Therefore, `gpu_row_blocks` was introduced to also split the result matrix along the row dimension. This makes it possible to adjust the block size so that each block fits into GPU memory for larger input sizes.

In first tests, the new GPU Hamming distance metric can run the largest full TCR dataset I currently have access to, containing around 8 million cells, in about 2 minutes on an NVIDIA A30 GPU. The same computation took around 35 minutes with the CPU Hamming implementation using 64 CPUs. Since this dataset is still too small for extensive stress testing, I also tested the `GPUHammingDistanceCalculator` directly with 25 million synthetically generated unique CDR3 sequences, which took around 30 minutes on an NVIDIA A30 GPU.

## Main changes

- Add row and column block splitting for GPU Hamming distance computation.
- Sort sequences within each row and column block by sequence length to group similar-length sequences for GPU computation.
- Skip unnecessary computations for symmetric distance matrices.
- Add better validation for GPU Hamming parameters.
- Update documentation for the GPU Hamming block parameters.
- Add tests for different block configurations.

## API change

`gpu_n_blocks` has been replaced by `gpu_col_blocks`. A new parameter, `gpu_row_blocks`, has been added.


src/scirpy/ir_dist/metrics.py
```diff
@@ -775,47 +792,60 @@ class GPUHammingDistanceCalculator(_MetricDistanceCalculator):
...
 
     def __init__(
         self,
         *,
         cutoff: int = 2,
-        gpu_n_blocks: int = 10,
-        gpu_block_width: int = 1000,
+        n_blocks: int = 1,
+        gpu_tile_rows: int = 100_000,
+        gpu_tile_cols: int = 100_000,
```
grst, line 830: Could you please keep the previous parameters and mark them as deprecated? 
It's fine if they do nothing and default values are used instead, I'd just like to avoid hard failures. 

You can add a dependency on our new util package [scverse-misc](https://github.com/scverse/scverse-misc) and use the [`@deprecated_arc`](https://scverse-misc.readthedocs.io/stable/generated/scverse_misc.deprecated_arg.html) decorator.

grst (commented):
Haven't reviewed the algorithm in detail, but the tests guarantee consistency with the reference, so this should be fine.

docs/tutorials/large-datasets.md
```diff
@@ -59,10 +59,47 @@ First, install the optional `cupy` dependency:
...
 ir.pp.ir_dist(mdata, metric="gpu_hamming")
 ```
 
-to take advantage of GPU acceleration.
+to use the first available GPU. The calculation is split into tiles. Their size can be adjusted with
+`gpu_tile_rows` and `gpu_tile_cols`: smaller tiles use less GPU memory but increase processing overhead.
+`gpu_tile_buffer_cols` controls the initially reserved space for retained distances and is enlarged automatically
+if necessary.
+
+### Using multiple GPUs
```
grst, line 73: Do you think linking to [the rapids-singlecell docs](https://rapids-singlecell.readthedocs.io/en/latest/out_of_core.html) for advanced usage would make sense here?

src/scirpy/ir_dist/metrics.py
grst: you can take a look at #735 that introduces scverse-misc for deprecations across the codebase.

felixpetschko:
Hi @grst! I have now deprecated the old parameters. If they are still used, the function sends a warning and falls back to the default values of the new parameters.

The new parameters work as follows: The computation of the resulting distance matrix is divided into tiles. `gpu_tile_rows` and `gpu_tile_cols` determine the size of these tiles in terms of rows and columns. `gpu_tile_buffer_cols` determines the width (number of columns) of the result buffer for each tile. This buffer has to fully fit into VRAM. It grows automatically when needed, provided enough VRAM is available. However, growing the buffer requires recomputing the tile whose results exceeded the previous, smaller buffer.

The advantage of the new parameters is that you only need to adapt the tile parameters to your GPU's VRAM and the expected result sparsity, without having to care about the total dataset size. Additionally, I think the default parameters should now be sufficient for most scenarios.

docs/tutorials/large-datasets.md
felixpetschko: Good idea, I linked it 👍

grst (approved):

PR #736, opened 2026-07-29
Clean up usage of _is_na, _is_true and _is_false

### LINKED ISSUE #380: clean up usage of _is_na, _is_true and _is_false
### Description of feature

Since #356, we have additional guarantees about the data types in `.obsm["airr"]` and there is no danger that AnnData changes the data type of a pandas column anymore during save/load. Therefore, we can get rid of most instances of `_is_na`, `_is_true`, and `_is_false` checks. The only remaining instances should be in the IO module sanitizing input before it is stored in AnnData. 

Cleaning this up should improve performance, and make the code more legible.

When doing so, tests should be added that everything still works as expected when creating `AirrCell` objects from data that does contain string representations of `NaN`/`True`/`False`/...

grst:
Since #356, the data types in `.obsm["airr"]` are guaranteed, so downstream
code no longer needs to guard against string representations of `NaN`,
`True` and `False`. Replace those checks with `pandas.isnull`/`notnull`,
which is both faster and easier to read. The only remaining occurrences are
in `scirpy.io`, where input data is sanitized before it is stored.

To make that guarantee hold for arbitrary input, `AirrCell.add_chain` now
also casts strings to the type declared in the AIRR rearrangement schema.
Previously, building `AirrCell` objects from data that spelled out booleans
or numbers as strings (e.g. `productive="True"`) produced an
`.obsm["airr"]` array with union types that `pp.index_chains` could not
process.

Two follow-ups fell out of this:

* `pl.vdj_usage` relied on `astype(str)` turning missing gene calls into
  the string `"nan"`, which no longer happens with pandas' nullable string
  dtypes. As a result the `none` segment was silently dropped from the
  plot. Missing values are now replaced with `"none"` explicitly.
* Two test fixtures spelled a missing clonotype as the string `"NaN"` in
  `obs`. Those now use `None`.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Closes #380

- [ ] CHANGELOG.md updated
- [ ] Tests added (For bug fixes or new features)
- [ ] Tutorial updated (if necessary)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
+## Unreleased
+
+### Fixes
+
+  - `AirrCell.add_chain` now casts values to the type declared in the AIRR rearrangement schema. Previously, building
+    `AirrCell` objects from data that spelled out booleans or numbers as strings (e.g. `productive="True"`) resulted in
+    an `adata.obsm["airr"]` array with union types that `pp.index_chains` could not process
+    ([#380](https://github.com/scverse/scirpy/issues/380)).
+  - `pl.vdj_usage` again shows cells without a gene call as a separate `none` segment instead of silently omitting them.
+
+### Chore
+
+  - Remove the `_is_na`/`_is_true`/`_is_false` checks outside of the `scirpy.io` module. Since scirpy guarantees the
+    data types in `adata.obsm["airr"]`, missing values can be detected with `pandas.isnull`, which is faster and easier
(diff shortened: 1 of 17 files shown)


Qinzi27:
Hi @grst, I reproduced the two `test_vdj_usage_missing_genes` failures locally with pandas 2.3.3. With the same other dependencies, they pass on pandas 3.0.5.

The cause is that `astype(str)` converts missing values to literal strings such as `"<NA>"` on pandas 2, so the following `fillna("none")` cannot replace them. Changing it to `astype("string").fillna("none")` fixes both failures.

On Windows/Python 3.12, all six existing VDJ plotting cases pass with that change on both pandas versions. I also checked the fixture's per-gene counts and normalized weights for AnnData/MuData; the missing group contains six cells, with 15 cells total.

Issue #731, opened 2026-07-27
Fix test warnings


grst:
There are a lot of warnings when running the tests. We should address or explicitly ignore them.

PR #750, opened 2026-09-16
Fix deprecated AnnData coordinate-key checks in clonotype networks

Qinzi27:
Plotting a clonotype network and exporting its layout to igraph emit an AnnData `obsm_keys()` FutureWarning. Check membership directly in `adata.obsm` in both paths, preserving the existing missing-coordinate errors.

Related to #731; this addresses one warning class and leaves other warnings visible.

The existing plotting and igraph tests now treat this specific warning as an error. The igraph tests retain their node-count and coordinate assertions and cover a custom basis. Additional tests check the missing-coordinate error for both APIs with AnnData and MuData and default and custom bases.

Validation on Windows with Python 3.12.14 and AnnData 0.13.3.post0:

- The same 140 regression cases fail on the original source with the `obsm_keys` FutureWarning and pass on the fixed source, with no skips. Both runs use the same interpreter and byte-identical test files.
- The full clonotype and plotting modules report 272 passed and 4 strict XPASS results. Those four cases also produce strict XPASS on the original source: their existing Windows xfail marks expect an igraph coordinate issue, but the assertions pass in this environment. No xfail settings were changed.
- `pip check`, Ruff lint/format checks on the changed Python files, and `git diff --check` pass.
- The full repository suite and Linux/Python version matrix were not run locally.

- [x] CHANGELOG.md updated
- [x] Regression tests added

No tutorial update is needed for this compatibility fix.

Implementation and test preparation were assisted by AI. The before/after test runs and baseline failure messages were checked explicitly.


src/scirpy/tests/test_clonotypes.py
```diff
@@ -321,8 +322,14 @@ def test_clonotype_network_mask_obs(adata_conn):
 
 
 @pytest.mark.extra
-def test_clonotype_network_igraph(adata_clonotype_network):
-    g, lo = ir.tl.clonotype_network_igraph(adata_clonotype_network)
+@pytest.mark.filterwarnings("error:.*obsm_keys.*:FutureWarning")
```
grst, line 325: Why is it necessary to add filterwarnings here? I thought the point of this PR was to fix this warning at its source?

Qinzi27: Thanks for reviewing. The two production changes fix the warning at its source by replacing obsm_keys() with membership checks on obsm.
I added the error marker as a regression check: the current test configuration otherwise allows this FutureWarning, so the old call could be reintroduced without failing these tests. I also checked that the replacement preserves the tested outputs and error behavior.

grst: Thanks, I wasn't aware how the syntax of filterwarnings works.

grst (approved):

